PR #83, opened 2014-12-23
Removes obsolete patches project

nakomis:


grkvlt:
@nakomis I actually want to keep the project, even if it's empty, so instead I've deleted the unwanted files in another commit

grkvlt:
Merged apache/incubator-brooklyn#364

PR #85, opened 2015-01-13
Docker and examples modules build bundles

sjcorbett:
See also https://github.com/brooklyncentral/advanced-networking/pull/20.

### DIFF
--- a/docker/pom.xml
+++ b/docker/pom.xml
@@ -92,4 +92,22 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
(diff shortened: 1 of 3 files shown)


grkvlt:
Looks good to merge. Do you not need to change the project type to `bundle` as well?

sjcorbett:
I think instructing maven-bundle-plugin to [include jars](https://github.com/brooklyncentral/clocker/pull/85/files#diff-600376dffeb79835ede4a0b285078036R194) is sufficient.

PR #89, opened 2015-01-22
Add deploy repo

neykov:
Based (but does not depend) on https://github.com/brooklyncentral/clocker/pull/88

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ services:
 
 ### Building from source
 
+<!-- CLOCKER_VERSION_BELOW -->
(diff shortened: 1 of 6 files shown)


grkvlt:
@neykov So I just need to merge this and since it includes #88 that should just be closed automatically, right?

neykov:
#88 won't be closed automatically. Better merge #88 first and after that this one.

grkvlt:
The commits have the same ID (235cdbf) so it should be OK, I think

grkvlt:
Yep, worked, both got closed. Nice!

neykov:
Great, didn't know this is supported :)

PR #90, opened 2015-01-27
Fix DockerHostSshDriver isRunning

sjcorbett:
Without `noExtraOutput` the `stdout` of the command when Docker is not running is like this:
```
/usr/sbin/service
docker stop/waiting
Executed /tmp/brooklyn-20150127-171334283-kshT-check-running_DockerHostImpl_i.sh, result 0
```
Which always contains 'running'!


grkvlt:
Oops! Thanks for this @sjcorbett I'll merge your change.

PR #93, opened 2015-01-30
Fixes up - mostly - integration and live tests

sjcorbett:
Fixes the integration test classes.

DockerInfrastructureLiveTest is most interesting. It runs in jclouds:softlayer:ams01 by default. Location specs can be injected by running with `-DlocationSpec=...`.


grkvlt:
Looks good, merging

PR #97, opened 2015-02-16
Run entities from named images

sjcorbett:
* Entities can be run from named images. 
* Adds an HAProxy entity

The following should work:
```yaml
services:
- type: brooklyn.entity.proxy.haproxy.HAProxyController
  brooklyn.config:
    imageName: haproxy
    imageTag: latest
    installDir: /usr/local/sbin/
    runDir: /usr/local/etc/haproxy/
```

See also `HAProxyIntegrationTest`.

This is submitted for review expecting that a few adjustments will be necessary. For one, I haven't altered `DockerUtils.imageName` yet.

### DIFF
--- a/docker/src/main/java/brooklyn/location/docker/DockerHostLocation.java
+++ b/docker/src/main/java/brooklyn/location/docker/DockerHostLocation.java
@@ -70,6 +70,7 @@ import brooklyn.util.ssh.BashCommands;
 import brooklyn.util.text.Strings;
 
 import com.google.common.base.Objects.ToStringHelper;
+import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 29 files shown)


sjcorbett:
The blueprint snippet given above should be the following instead:
```yaml
          docker.image.name: haproxy
          docker.image.tag: latest
          install.dir: /usr/local/sbin/
          run.dir: /usr/local/etc/haproxy/
```

grkvlt:
Thanks, this looks great @sjcorbett, I'll review and let you know if there are any issues

examples/src/main/assembly/files/blueprints/dockerfile-mysql.yaml
```diff
@@ -18,7 +18,7 @@ origin: "https://registry.hub.docker.com/_/mysql/"
 locations:
 - my-docker-cloud
 services:
-- serviceType: brooklyn.entity.container.docker.application.DockerfileApplication
+- serviceType: brooklyn.entity.container.docker.application.VanillaDockerApplicationion
```
grkvlt: `VanillaDockerApplication` and should be `type:` as well

docker/src/main/java/brooklyn/location/docker/DockerHostLocation.java
```diff
@@ -156,24 +159,33 @@ public DockerContainerLocation obtain(Map<?,?> flags) throws NoMachinesAvailable
...
+            final String fullyQualifiedImageName = Os.mergePaths(repository, imageName);
 
             // Lookup image ID or build new image from Dockerfile
-            LOG.warn("ImageName for entity {}: {}", entity, imageName);
-            String imageList = dockerHost.runDockerCommand("images --no-trunc " + Os.mergePaths(repository, imageName));
-            if (Strings.containsLiteral(imageList, imageName)) {
-                // Wait until committed before continuing
+            LOG.info("ImageName for entity {}: {}", entity, imageName);
+
+            if (dockerHost.getImageNamed(fullyQualifiedImageName, imageTag).isPresent()) {
```
grkvlt: We should save the `Optional` here since we repeat this method call just to call `get()` on it a couple of lines later

examples/src/main/assembly/files/blueprints/dockerfile-mysql.yaml
grkvlt: Will fix that post merge, though

PR #99, opened 2015-03-23
fix riak webapp with weave blueprint

andreaturli:


grkvlt:
Thanks @andreaturli this fixes the issue caused by https://github.com/docker-library/haproxy/issues/3

grkvlt:
I'll update to a newer version tag once the Docker image fix is pushed

PR #101, opened 2015-04-10
Fix build, use the correct return type

neykov:
Prompted by destiny on IRC that the build fails, confirmed on my side.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@
 
 language: java
 jdk:
-- openjdk7
+- openjdk6
 script: mvn clean install
 before_deploy: "mkdir artifacts && mv examples/target/brooklyn-clocker-dist.tar.gz artifacts/brooklyn-clocker-${TRAVIS_BUILD_NUMBER}-${TRAVIS_COMMIT}-dist.tar.gz"
 deploy:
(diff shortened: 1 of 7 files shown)


andreaturli:
@neykov this should be already fixed on [jclouds/jclouds-labs](https://github.com/jclouds/jclouds-labs/blob/master/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java) not sure why we are propagating this error here

neykov:
So the correct fix would be to remove `DockerTemplateOptions` altogether from clocker project?

grkvlt:
Yeah, I will remove once we have the next jclouds release. I have fixed those option return values in another commit too.

@neykov or @andreaturli I believe they are actually broken in jclouds-labs though, still, so check and issue a PR there if necessary.

andreaturli:
@grkvlt right! so I've issued a PR https://github.com/jclouds/jclouds-labs/pull/164
could you please review so I can merge it?

grkvlt:
Looks fine. Note that you could also at the same time remove the casting from the `Builder` static methods, as it isn't needed. So the static method for `env` becomes:

```Java
/**
 * @see DockerTemplateOptions#env(Iterable)
 */
public static DockerTemplateOptions env(Iterable<String> env) {
   DockerTemplateOptions options = new DockerTemplateOptions();
   return options.env(env);
}
```

And the same for all the other options that are defined in this class.

PR #103, opened 2015-04-16
Keep the last docker host when its last docker container is removed

mattunderscorechampion:
When DockerInfrastructure.REMOVE_EMPTY_DOCKER_HOSTS is set to true and a container is removed a check will be performed to see if any containers are running on the host. If the host has no other containers it will also be removed. This will also remove the last host when the last container is removed. I would suggest that the last host should be kept even when it is empty.


grkvlt:
@mattunderscorechampion looks good, I'll merge this - we should make the scale-down mechanics more configutable, but can do this later...

PR #109, opened 2015-06-02
update POM to skip testing for webUI (as there are no tests)

csabapalfi:
Maven gets confused (because of parent level test excludes) and fails with the error below unless modules without test have tests skipped.

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.17:test (default-test) on project brooklyn-clocker-console: groups/excludedGroups require TestNG or JUnit48+ on project test classpath -> [Help 1]```

Should be reverted as soon as web UI will have tests added.

### DIFF
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -59,6 +59,13 @@
                     </webResources>
                 </configuration>
             </plugin>
+            <plugin>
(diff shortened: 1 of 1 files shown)


csabapalfi:
Superseded by #110.

PR #110, opened 2015-06-02
Fix clocker build

csabapalfi:
This is to fix clocker builds:

1. update POM to skip tests for web console (#109) 
2. some more type casts to accomodate recent brooklyn changes

### DIFF
--- a/docker/src/main/java/brooklyn/entity/container/docker/DockerContainerImpl.java
+++ b/docker/src/main/java/brooklyn/entity/container/docker/DockerContainerImpl.java
@@ -68,7 +68,7 @@ import brooklyn.location.docker.DockerHostLocation;
 import brooklyn.location.dynamic.DynamicLocation;
 import brooklyn.location.jclouds.JcloudsLocation;
 import brooklyn.location.jclouds.JcloudsLocationConfig;
-import brooklyn.location.jclouds.JcloudsSshMachineLocation;
+import brooklyn.location.jclouds.JcloudsMachineLocation;
 import brooklyn.location.jclouds.templates.PortableTemplateBuilder;
 import brooklyn.management.LocationManager;
 import brooklyn.networking.portforwarding.subnet.JcloudsPortforwardingSubnetLocation;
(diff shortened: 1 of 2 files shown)


docker/src/main/java/brooklyn/entity/container/docker/DockerContainerImpl.java
```diff
@@ -450,7 +450,7 @@ public DockerContainerLocation createLocation(Map flags) {
 
         try {
             // Create a new container using jclouds Docker driver
-            JcloudsSshMachineLocation container = host.getJcloudsLocation().obtain(dockerFlags);
+            JcloudsMachineLocation container = (JcloudsMachineLocation) host.getJcloudsLocation().obtain(dockerFlags);
```
grkvlt: Keep cast but change type to `JcloudsSshMachineLocation` as we know this is what will be returned from the jclouds Docker driver at present.

docker/src/main/java/brooklyn/entity/container/docker/DockerContainerImpl.java
```diff
@@ -491,7 +491,7 @@ public DockerContainerLocation createLocation(Map flags) {
                     .configure(flags)
                     .configure(DynamicLocation.OWNER, this)
                     .configure("machine", container) // the underlying JcloudsLocation
-                    .configure(container.config().getBag().getAllConfig())
+                    .configure(((ConfigurationSupportInternal) container.config()).getBag().getAllConfig())
```
grkvlt: Not convinced we need this cast, type returned by `ontainer.config()` should be `ConfigurationSupportInternal` see https://github.com/apache/incubator-brooklyn/blob/master/core/src/main/java/brooklyn/location/basic/AbstractLocation.java#L350

csabapalfi:
You're right. Had outdated SNAPSHOTs and my local builds were failing without that cast. Removed it now.

grkvlt:
Looks fine now, merging :frog:

PR #111, opened 2015-06-03
Add simple initial Dockerfile

csabapalfi:
This Dockerfile allows building a clocker image once you build the dist tar.gz locally.

Once built you can just use `docker run -d -p 8082:8081 -v ~/.brooklyn:/root/.brooklyn clocker` to start clocker.

Few details:
* I assume we want a docker build (and push to docker-hub) as part of our build instead.
* The image is built on top of the super-light alpine distro hence the image size is relatively small. 
* brooklyn config has to be bind mounted to the clocker container to work

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+console
+docker
+patches
+examples/target/brooklyn-clocker-dist
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM gliderlabs/alpine:3.1
+
+RUN apk-install openjdk7-jre-base
+RUN apk-install bash
+
+ADD examples/target/brooklyn-clocker-dist.tar.gz .
+
+WORKDIR brooklyn-clocker
+
+VOLUME /root/.brooklyn
+
+EXPOSE 8081
+
+CMD bin/clocker.sh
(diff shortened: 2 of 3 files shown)


andreaturli:
really like the idea of using alpine distro

andreaturli:
this looks good to me. A simple README file would be a nice-to-have

andreaturli:
@csabapalfi would it make sense to have clocker running on Docker which is able to provision the Docker Cloud on the Docker Engine where Clocker container is running?

csabapalfi:
Thanks @andreaturli. Yes, why not :) Or maybe I'd rather have my clocker containers running separately as part of management infrastructure. I'm guessing that's how most people use brooklyn as well. 

Still in theory nothing is stopping you from doing that apart from the fact that clocker is able to replace failed docker hosts/containers but it can't do this for itself and requires a bit more attention than your service/app containers managed by it.

I'm new on the project so @grkvlt might have a better view on this :)

csabapalfi:
Added a docker-compose.yml, too so now it's even easier to start clocker if you have docker-compose installed: `docker-compose run clocker <location> <network>`

grkvlt:
We can update the README separately, I think. I'll merge this now so more people can test it. Cheers :frog:

PR #112, opened 2015-06-03
initial Vagrantfile

csabapalfi:
This is a basic Vagrantfile which sets up port forwarding and a synced directory for brooklyn config.

Clocker is not (yet) provisioned as a service within the VM. It can be started by `vagrant ssh` and manually invoking the clocker.sh script for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ obj
 *.swp
 .DS_Store
 *.swp
+
+.vagrant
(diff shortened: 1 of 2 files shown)


grkvlt:
Also good. Merging.

grkvlt:
Again, leaving documentation updates as a separate task for now. @csabapalfi perhaps create a wiki page with some info on this and the Docker file though?

PR #114, opened 2015-06-16
Make sure node app environment variables are resolved

csabapalfi:

PR #117, opened 2015-06-18
fix container start effector call

csabapalfi:
#114 breaks container start as `invokeEffectorWithArgs` expect a list and not a map. Changing to `invokeEffector` fixes this.


grkvlt:
@csabapalfi my bad, thanks for the fix!

grkvlt:
FYI, this will be obsoleted by a new `ContainerExecutionContext` class that manages effector execution and variable interpolation for entities that are hosted in a `DockerContainer` such as `VanillaDockerApplication` - pull request to follow...

PR #121, opened 2015-06-23
Return VanillaDockerApplication context for containers

grkvlt:


grkvlt:
This is wrong. The problemn is that you cannot reference another application in a blueprint. I think this needs to be a Brooklyn change, with a scope like `APPLICATION` that would look up an app by name or id, and then `component()` or `entity()` can be used to select an entity from that. Also, the `GLOBAL` scope is misleading, but probably too late to change. Perhaps a `BROOKLYN` scope that returns _all_ entities in _all_ applications, for selectioin?

PR #118, opened 2015-06-18
Support docker exec (instead of SSH)

csabapalfi:
This is still work in progress. It's a naive implementation for using docker exec instead of SSHing in to container to execute commands.

* need to rebase this on top of the latest master (with latest weave) then force push
* env vars are not handled properly
* there are a few places where DockerContainerLocation is still used as an SSH location as it implements that interface (e.g. MachineLifecycleEffectorTasks.resolveOnBoxDir)

PR #130, opened 2015-06-29
Support docker exec (instead of SSH)

grkvlt:
Minor updates to #118 to allow it to be merged for testing


grkvlt:
@csabapalfi I'd like to see a bit more testing on this before we merge

PR #132, opened 2015-07-01
change to 0.8.3-SNAPSHOT in branch 0.8.x

aledsage:


aledsage:
@grkvlt Travis is still failing with:

    Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile (default-compile) on project brooklyn-clocker-docker: Fatal error compiling: invalid target release: 1.7 

However, it passes locally. Could this be an issue with the Travis configuration?

andreaturli:
it's weird because although .travis.yaml asks for jdk7 I can see on the first line of the log

```
Switching to OpenJDK6 (java-1.6.0-openjdk-amd64), JAVA_HOME will be set to /usr/lib/jvm/java-6-openjdk-amd64
$ java -Xmx32m -version
java version "1.6.0_34"
```
which makes maven-compiler unhappy I think

andreaturli:
Comparing the [doc](http://docs.travis-ci.com/user/languages/java/#Testing-Against-Multiple-JDKs) with https://github.com/brooklyncentral/clocker/blob/master/.travis.yml#L17 looks like the indentation is wrong.

If not specified, travis will use oracleJDK7 maybe we should try that

grkvlt:
Committing change to version in this release branch *only* to allow external scripts to run.

PR #136, opened 2015-07-03
Fix handling of TLS keys and certificates

grkvlt:
Docker 1.7.0 requires changes to the way we configure TLS. This patch creates the server and client certificates and keys automatically for each Docker host, with a shared CA certificate on the Clocker server.


grkvlt:
@csabapalfi Currently the Java SSL trust management does not seem to work, causing certificate errors when accessing the Docker REST API, although the generated certificates work when used from the command line.

csabapalfi:
TLS keys are now fine but docker is now complaining that it can't parse the `devices` field in the JSON we send as an object (it expects an array). Fixing that now.

grkvlt:
@csabapalfi it's probably in the DTOs so to change them you need to modify jclouds, and rebuild that. Because of the way the `AutoValue` DTOs are built with Maven, you need to make a new jclouds release. I can help with that if you let me know the changes needed. I'll make a `1.9.1-clocker-pre.2` version.

csabapalfi:
@grkvlt for now I just patched `DockerComputeServiceAdapter` to set the `devices` field to null when building the container config since we don't use that field at all. (Otherwise it's initialised as an empty map and causes the error)

Once I get clocker working then I fix the devices field properly and submit a PR to jclouds. Same goes for the TLS fixes.

grkvlt:
@csabapalfi good call, OK. We can patch jclouds later.

csabapalfi:
It took me a while to figure out why the TLS stuff was broken. First I thought my fix is wrong but turned out jclouds was overwriting my SSLContext :rage1: Anyways I hacked my way around that now and a proper fix is on it's way.

csabapalfi:
In the end I decided to work on jclouds directly as there were more and more changes required to make docker 1.7 work. See https://github.com/jclouds/jclouds-labs/pull/188

Now I'm porting the custom SSLContext fix (instead of having to use the `keytool` as jclouds currently works) then we can build `jclouds-labs-1.9.1-clocker-pre.2`

csabapalfi:
Removed all the local jclouds patches. Testing your pre-release brooklyn/jclouds version commit now.

grkvlt:
I am merging this, and will do a further update to another jclouds pre-release if nescessary, but I think we can work with the 1.9.1-clocker-pre.2 for now

PR #138, opened 2015-07-22
Fix for setting VLAN id in SoftLayer

grkvlt:
If a VLAN is configured in the blueprint or location, via the `SdnProvider#VLAN_ID` config key or the `primaryBackendNetworkComponentNetworkVlanId` template option, Clocker will use it, otherwise it will accept whatever is provided by SoftLayer. It then sets the `SdnProvider#VLAN_ID` attribute on the SDN provider entity, using the VLAN id for the first host. This is what will then be used for all subsequent hosts, no matter whether it was assigned by SoftLayer or configured in the blueprint.

Depends on #136

### DIFF
--- a/docker/src/main/java/brooklyn/entity/container/docker/DockerHostImpl.java
+++ b/docker/src/main/java/brooklyn/entity/container/docker/DockerHostImpl.java
@@ -28,7 +28,10 @@ import org.jclouds.compute.config.ComputeServiceProperties;
 import org.jclouds.compute.domain.OsFamily;
 import org.jclouds.compute.domain.TemplateBuilder;
 import org.jclouds.compute.domain.Volume;
+import org.jclouds.softlayer.SoftLayerApi;
 import org.jclouds.softlayer.compute.options.SoftLayerTemplateOptions;
+import org.jclouds.softlayer.domain.VirtualGuest;
+import org.jclouds.softlayer.features.VirtualGuestApi;
 import org.jclouds.softlayer.reference.SoftLayerConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


grkvlt:
Looks good from testing, for me on SoftLayer

docker/src/main/java/brooklyn/entity/container/docker/DockerHostImpl.java
```diff
@@ -481,6 +506,25 @@ public void deleteLocation() {
     protected void preStart() {
         getDriver().configureSecurityGroups();
 
+        // Save the VLAN id for this machine
+        MachineProvisioningLocation location = getInfrastructure().getDynamicLocation().getProvisioner();
+        if (isJcloudsLocation(location, SoftLayerConstants.SOFTLAYER_PROVIDER_NAME)) {
```
ahgittin: a bit naughty to do this here!  can it go into `brooklyn-jclouds` project as a `SoftLayerSameVlanLocationCustomizer` ?

i think it would:

0. if vlan properties are set explicitly on the location, return
1. set `scope` = value of a location config `"SoftLayerSameVlanLocationCustomizer.scopeUid"` defined on the location 
2. in a `synchronized (SoftLayerSameVlanLocationCustomizer.class)` block (to prevent multiple threads creating the semaphore) get the location config `"SoftLayerSameVlanLocationCustomizer.scopeSemaphores."+scope`, or if not present set that to be a semaphore with one lock available
3. acquire the lock on the semaphore, blocking (calling `Tasks.setBlockingDetails("waiting on VLAN info")` before and `resetBlockingDetails` after)
4. look do we have any VM's in any machine provisioning location with the same scope, if so take the VLAN ID's from that location and set them on the provisioning request
5. return, without creating VM, then ensure we get a callback on creation *or on failure* and release the lock at that point

this should do the right thing, ie ensuring only one thread can create a VM if we don't yet have vlan details, and getting existing vlan details otherwise

/cc @aledsage the problem this solves is that SL may put your VM's into 

also update the `more-on-clouds.md` page in brooklyn docs to refer to this

ps @grkvlt this is an awesome idea i've nicked from you :)

grkvlt: Sadly, there's no callback mechanism for failures on the `JcloudsLocationCustomizer` so we probably just have to wait on a lock with timeout. The way the customizer works is there are multiple methods that can be implemented, the ones that interest us are the post-creation call, which passes in a created machine, and the pre-creation call which offers us a `TemplateOptions` to modify. I think I can work with these. I'm also going to default the scope to the application ID if nothing is defined on the location, which seems like the way most people want this to work.

PR #141, opened 2015-08-10
Add alpine support when layering sshd on top

csabapalfi:
This change makes it possible to layer sshd on top of an alpine image.

e.g https://github.com/csabapalfi/hello-alpine-node can now be deployed with clocker.

I tested locally and using the SshdDockerfile results in an SSHable alpine image.


csabapalfi:
`onbox.base.dir.skipResolution: true` is required (and potentially some more changes around runtime directory setup)

PR #146, opened 2015-08-16
Add alpine support when layering sshd on top

grkvlt:
Signed-off-by: Csaba Palfi <csaba@palfi.me>


grkvlt:
Updated version of #141 rebased with `master`

PR #147, opened 2015-08-19
enhance MicroServiceImage

andreaturli:
- if the my-docker-cloud is not available, it will first install a DockerInfrastructure with a single DockerHost

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ obj
 *.swp
 
 .vagrant
+brooklyn-clocker/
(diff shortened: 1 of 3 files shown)


andreaturli:
@grkvlt I think it is failing because last available brooklyn 0.8.0-SNAPSHOT is from Aug, 10.

grkvlt:
@andreaturli yes, I'll fix the imports and rebase, see if that fixes things

grkvlt:
See #148 for rebased version

PR #153, opened 2015-09-10
Classnames

sjcorbett:


grkvlt:
@sjcorbett thanks - the refactor has been difficult to keep track of, hopefully we have caught up now

PR #154, opened 2015-09-12
Update to the new brooklyn version with renamed packages

bostko:


grkvlt:
@bostko I changed Clocker to use `0.8.0-incubating` for the moment, as this is the current 'stable' version.

grkvlt:
@bostko @aledsage @ahgittin FYI there was *no* `0.8.0-incubating` branch for `brooklyncentral/advanced-networking` for some reason, so I am creating one and deploying to Sonatype to make Clocker work with this Brooklyn release.

grkvlt:
@bostko there's another issue with changing to 0.9.0 for Clocker, which I only just noticed. The way catalog entries are parsed and created has changed, particularly with regard to appearance in the _Add Application_ wizard. We will need to change the BOM and YAML files, as well as reviewing the use of `@SetFromFlag` and `@Catalog`/`@CatalogConfig` annotations in entities. Do you have any advice?

ahgittin:
Re parsing catalog entries I think we tried to maintain backwards compatibility so if things aren't working w 0.9.0 it may be that we want to tweak Brooklyn rather than Clocker. /cc @neykov 

What are the issues @grkvlt ?

grkvlt:
@ahgittin @bostko Now *all* entries marked `@SetFromFlag` show up in the catalog wizard for `DockerInfrastructure` itself (an `Application`) and any derived catalog entries (templates) from the BOM show no config options, like this:

```YAML
 - id: docker-cloud
    itemType: template
    item:
      services:
      - type: docker-infrastructure
```

ahgittin:
ah it's the config keys work that @neykov has done.

is it worth extra steps to support backwards compatibility here?  ie if a BOM yaml does not include the "brooklyn.parameters" section then it inherits from the targetted class.  and if a class has `@CatalogConfig` annotations then *only* those are taken as parameters when generating a catalog item?

alternatively we could deprecate `@CatalogConfig` and put catalog config in YAML, ie a simple update to the BOM's you have.

grkvlt:
@ahgittin OK so is there an example of the `brooklyn.parameters` config somewhere we can use to modify our BOM files?

ahgittin:
see the `catalog/index.md` file in the doc, and/or https://github.com/apache/incubator-brooklyn/pull/1001

do you have a view on backwards compatibility?

@aledsage are canopy using the `@CatalogConfig` annotations?  if so then we will need it in any case.

grkvlt:
I'm happy to make the breaking change since backwards compatibility is probably going cause precedence problems, better to have a single _One True Way_ of doing things. @Graeme-Miller we will need to update the Clocker catalog BOM files and applications. See issue #191

neykov:
@ahgittin @grkvlt `@Catalog`/`@CatalogConfig` should work out of the box. Let me check what's causing this.

neykov:
Can confirm this is because of the parameters changes. Brooklyn is not able to parse the catalog items when initializing, just stuffing them in without knowing the full details of the items. This causes the config key list to be empty. Will fix.

My expectation was that this would happen only when installing actually broken items (for example items containing placeholder text that will work only after replacing). Turns out this is not the case.

neykov:
@grkvlt `@CatalogConfig` fixed in https://github.com/apache/incubator-brooklyn/pull/1036

grkvlt:
Obsoleted by #201

PR #157, opened 2015-09-15
Forbid imports from com.google.api.client.repackaged

sjcorbett:


grkvlt:
@sjcorbett Nice! I do this accidentally a lot... ;(

PR #163, opened 2015-09-22
Fix VanillaDockerApplication

Graeme-Miller:
Moved configuration from DockerServiceTypeResolver to VanillaDockerApplication, so that VanillaDockerApplication can be used


docker/src/main/java/brooklyn/entity/container/docker/application/VanillaDockerApplication.java
```diff
@@ -70,6 +70,12 @@
     @SetFromFlag("volumes")
     ConfigKey<List<String>> DOCKER_CONTAINER_VOLUME_EXPORT = DockerAttributes.DOCKER_CONTAINER_VOLUME_EXPORT;
 
+    @SetFromFlag("skipEntityStart")
```
grkvlt: Do we need this annotation? We won't actually be setting these ever, they're just there so that we can plug in alternate defaults. Same with next one.

grkvlt:
@Graeme-Miller can you rebase against master, after updating annotations, so that this can be merged pls.

PR #164, opened 2015-09-30
Fix dnsSearch type

googlielmo:
- dnsSearch declared as String, corrected to List<String>

See https://docs.docker.com/reference/api/docker_remote_api_v1.16/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ obj
 
 .vagrant
 brooklyn-clocker/
+
+.checkstyle
+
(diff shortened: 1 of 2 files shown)


grkvlt:
@googlielmo This should also be patched in `jclouds/jclouds-labs` which is the source of this file. I'm happy to merge into Clocker as well if you do this.

googlielmo:
Yes @grkvlt, it should be done there -- following @andreaturli 's advice I added a comment in the related https://issues.apache.org/jira/browse/JCLOUDS-900 as a reminder

googlielmo:
@grkvlt I opened https://github.com/jclouds/jclouds-labs/pull/207 in `jclouds/jclouds-lab`

grkvlt:
Thanks @googlielmo looks like that has beenmerged. I'm brining this into Clocker temporarily, until these changes make the next jclouds release, probably 1.9.2

PR #166, opened 2015-10-02
Added check that the SDN provider is not null

mattunderscorechampion:
I found a null pointer exception when using Softlayer, no SDN provider and more than one host. After the first host it attempts to lookup the VLAN ID from the SDN provider. The attributeWhenReady method throws a null pointer exception when the entity is null.

Added that the SDN provider is not null before attempting to lookup the VLAN ID.


grkvlt:
Thanks @mattunderscorechampion

PR #167, opened 2015-10-06
Initial Mesos integration

grkvlt:
Some partial Mesos and Marathon integration with a running Mesos cluster.


grkvlt:
Replaced by #185 with proper Marathon integration

PR #172, opened 2015-10-13
Adding configurable Docker registry

Graeme-Miller:
Updated version of this [pull request](https://github.com/brooklyncentral/clocker/pull/168) with a slightly better branch name.


docker/src/main/java/brooklyn/entity/container/docker/DockerContainerImpl.java
```diff
@@ -28,6 +28,7 @@
 
 import javax.annotation.Nullable;
 
+import com.google.common.collect.*;
```
grkvlt: Don't use wildcards for imports - should be configurable in IDE?

docker/src/main/java/brooklyn/entity/container/docker/DockerHostImpl.java
```diff
@@ -692,6 +701,31 @@ public void postStart() {
...
+                if(!externalRegistriesLoginDetails.containsKey(username)){
+                    throw new IllegalArgumentException(String.format("registry {%s} and username {%s} supplied but no password in DOCKER_EXTERNAL_REGISTRY_LOGIN_DETAILS", registry, username));
+                }
+
+                String password = externalRegistriesLoginDetails.get(username);
+                if(password == null) {
+                    throw new IllegalArgumentException(String.format("registry {%s} and username {%s} supplied but password in DOCKER_EXTERNAL_REGISTRY_LOGIN_DETAILS is null", registry, username));
+                }
+
+                runDockerCommand(String.format("login  -e \"fake\" -u %s -p %s %s", username, password, registry));
```
grkvlt: Should we maybe make email configurable?

grkvlt:
Should squash these into a single commit (or a few if it makes more sense) before merging. The formatting also needs updated, and can be done at the same time.

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructureImpl.java
```diff
@@ -427,6 +421,32 @@ public void stop() {
...
+            EntitySpec<DockerRegistry> spec = EntitySpec.create(DockerRegistry.class)
+                    .configure(DockerRegistry.DOCKER_HOST, firstEntity);
+
+            //mount volume with images stored on it
+            DockerRegistry dockerRegistry = addChild(spec);
+            Entities.manage(dockerRegistry);
+
+            LOG.debug("Starting a new Docker Registry with spec {}", spec);
+            dockerRegistry.start(ImmutableList.of(firstEntity.getDynamicLocation()));
+
```
grkvlt: Better to use `Entities#Start(Coillection)` for this, which will manage the entity and call start in a new `Task`.

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistry.java
```diff
@@ -0,0 +1,48 @@
...
+    AttributeSensorAndConfigKey<Entity, Entity> DOCKER_HOST = DockerContainer.DOCKER_HOST;
+
+    @SetFromFlag("env")
+    ConfigKey<Map<String, Object>> DOCKER_CONTAINER_ENVIRONMENT = ConfigKeys.newConfigKeyWithDefault(DockerContainer.DOCKER_CONTAINER_ENVIRONMENT.getConfigKey(),
+       ImmutableMap.<String, Object>of(
+                            "REGISTRY_HTTP_TLS_CERTIFICATE", "/certs/repo-cert.pem",
+                            "REGISTRY_HTTP_TLS_KEY", "/certs/repo-key.pem"
+            ));
+
+    @SetFromFlag("dockerHost")
```
grkvlt: Not useful on `AttributeSensor` fields

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistry.java
```diff
@@ -0,0 +1,48 @@
...
+    AttributeSensor<String[]> DOCKER_REGISTRY_CATALOG = Sensors.newSensor(new TypeToken<String[]>(){}, "docker.registry.catalog", "The docker registry catalog, which lists all of the available repositories");
```
grkvlt: Prefer typed `List<String>` to arrays

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.brooklyn.util.time.Duration;
+
+import brooklyn.entity.container.DockerAttributes;
+import brooklyn.entity.container.docker.DockerHost;
+import brooklyn.entity.container.docker.DockerInfrastructure;
+import brooklyn.entity.container.docker.application.VanillaDockerApplicationImpl;
+
+public class DockerRegistryImpl extends VanillaDockerApplicationImpl implements DockerRegistry {
+
+    private volatile HttpFeed httpFeed;
```
grkvlt: Should be `transient` to indicate ignore during serialisation. The `volatile` modifier is for concurrency. See this answer on [stackoverflow](http://stackoverflow.com/questions/3544919/what-are-transient-and-volatile-modifiers) for a better description.

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
```diff
@@ -0,0 +1,109 @@
...
+    private String getSSHMachineInstallDir() {
+        return config().get(DockerRegistry.DOCKER_HOST).sensors().get(SoftwareProcess.INSTALL_DIR);
+    }
+
+    @Override
+    public void connectSensors() {
+        super.connectSensors();
+        connectServiceUpIsRunning();
+
+        HostAndPort hostAndPort = BrooklynAccessUtils.getBrooklynAccessibleAddress(this, this.getConfig(DockerRegistry.DOCKER_REGISTRY_PORT));
```
grkvlt: Use `config().get(...)` instead of `this.getConfig(...)`

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        httpFeed = HttpFeed.builder()
+                .entity(this)
+                .period(Duration.seconds(3))
+                .baseUri(getAttribute(Attributes.MAIN_URI))
+                .poll(new HttpPollConfig<Boolean>(Attributes.SERVICE_UP)
+                        .onSuccess(Functions.constant(true))
+                        .onFailureOrException(Functions.constant(false)))
+                .poll(new HttpPollConfig<String[]>(DockerRegistry.DOCKER_REGISTRY_CATALOG)
+                        .suburl("/_catalog")
+                        .onSuccess(HttpValueFunctions.jsonContents("repositories", String[].class)))
```
grkvlt: This will need updated when the attribute type is changed to a `List`

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        String installDir = host.sensors().get(SoftwareProcess.INSTALL_DIR);
+        config().set(DockerAttributes.DOCKER_HOST_VOLUME_MAPPING, MutableMap.of(Os.mergePaths(installDir, "certs"), "/certs"));
+
+        // set location customizer to new inner class
+        // customizer does the SshMachineLocation stuff below
+        // is passed a DockerHostLocation, can get to SshMachineLocation from there
+
+
+        SshMachineLocation sshMachine = host.getDynamicLocation().getMachine();
+        sshMachine.installTo("classpath://brooklyn/entity/container/docker/registry/docker-registry-create-certs.sh", Os.mergePaths(getSSHMachineInstallDir(), "docker-registry-create-certs.sh"));
```
grkvlt: Shove `getSSHMachineInstallDir()` in a local variable (`dockerInstallDir`) and make `docker-registry-create-certs.sh` a constant as well

docker/src/main/java/brooklyn/location/docker/DockerHostLocation.java
```diff
@@ -159,8 +161,11 @@ public DockerContainerLocation obtain(Map<?,?> flags) throws NoMachinesAvailable
 
             Optional<String> baseImage = Optional.fromNullable(entity.config().get(DockerAttributes.DOCKER_IMAGE_NAME));
             String imageTag = Optional.fromNullable(entity.config().get(DockerAttributes.DOCKER_IMAGE_TAG)).or("latest");
+            Optional<String> imageRepo = Optional
```
grkvlt: Don't split a line after the classname

docker/src/main/java/brooklyn/entity/container/docker/DockerContainerImpl.java
Graeme-Miller: hmm, must have made this commit before I changed the IDE settings. Will go through all classes and check when we are ready to merge

docker/src/main/java/brooklyn/location/docker/DockerHostLocation.java
```diff
@@ -177,13 +182,23 @@ public DockerContainerLocation obtain(Map<?,?> flags) throws NoMachinesAvailable
...
             } else if (baseImage.isPresent()) {
+                String fullyQualifiedName = "";
+                String repoAndName = "";
+
+                if(imageRepo.isPresent()){
+                    fullyQualifiedName = imageRepo.get()+"/";
+                    repoAndName = imageRepo.get()+"/";
+                }
+                fullyQualifiedName += baseImage.get() + ":" +imageTag;
+                repoAndName +=  baseImage.get();
```
grkvlt:
```Java
String repoAndName = Joiner.on('/').join(Optional.presentInstances(ImmutableList.of(imageRepo, baseImage)));
String fullyQualifiedName = repoAndName + ":" + imageTag;
```
:frog:

docker/src/main/java/brooklyn/entity/container/docker/DockerHostImpl.java
Graeme-Miller: Maybe. Trouble is, we don't really need it here. Email is *usually* only used by the login command when registering a new user. When we use login here, we only really want to login, not register. 

The annoying thing is you still have to provide a non null email address. I suppose I could create an email config option with "fake" as the default?

There is an open git issue about this for docker [here](https://github.com/docker/docker/issues/6400)

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistry.java
Graeme-Miller: Me too, but it wouldn't work with List<String>
The JSON parser that populates this needs an array rather than a List

patches/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
```diff
@@ -24,6 +24,7 @@
 
 import org.jclouds.compute.ComputeService;
 import org.jclouds.compute.options.TemplateOptions;
+import org.jclouds.docker.internal.NullSafeCopies;
```
grkvlt: Definitely keep these files (the `jclouds-labs` PR in `patches`) as a separate commit

docker/src/main/java/brooklyn/entity/container/docker/DockerHostImpl.java
grkvlt: Meh. Change it to `fake@example.org` then, to make it clear it's an email address, and an unusable one, at thet.

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistry.java
grkvlt: You just need to iterate through the JSON objects. I'll paste an example at the relevant comment.

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
Graeme-Miller: I marked this as  volatile as I was following the precedent set in other classes, for example, RiakNodeImpl. Can remove if you think there will be no multi-threading issues?

Not sure about adding transient, is that something you would recommend? When would this be serialised? Possibly as part of persistence?

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
grkvlt: This example should work?

```Java
                .poll(new HttpPollConfig<List<String>>(DOCKER_REGISTRY_CATALOG)
                        .suburl("/_catalog")
                        .onSuccess(Functionals.chain(
                                HttpValueFunctions.jsonContents(),
                                JsonFunctions.walk("repositories"),
                                JsonFunctions.forEach(JsonFunctions.cast(String.class)))
                        .onFailureOrException(Functions.constant(Arrays.asList(new String[0]))))
```

grkvlt:
We need to add a catalog item to `default.catalog.bom` and `clocker.bom` for the registry and update or add example blueprints with the registry configured for both calico and weave. It would also be good to have an example that used the public `clockercentral/clocker` repository at [quay.io](https://quay.io/repository/clockercentral/clocker)

grkvlt:
@Graeme-Miller Looks good so far. Make sure the comments are addressed and some examples created. The commits need to be rebased against `master` and then squashed and formatted as well. I think it just needs a bit more testing and we should be good to merge.

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
grkvlt: Hm, probably cargo-cult copying going on there. There will never be multiple threads trying to read the reference to the feed. But, during persistance it may be serialized, and it has anonymous classes inside, so we don't want to do that. Really, `transient` is right, `volatile` is wrong and should be fixed in brooklyn - i'll review the code for that...

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructureImpl.java
```diff
@@ -23,15 +23,15 @@
...
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.annotation.Nullable;
 import javax.net.ssl.X509TrustManager;
 
+import brooklyn.entity.container.docker.registry.DockerRegistry;
+import org.apache.brooklyn.core.entity.AbstractApplication;
+import org.apache.brooklyn.entity.group.*;
```
grkvlt: No wildcard imports.

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructureImpl.java
```diff
@@ -23,15 +23,15 @@
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
grkvlt: No wildcard imports.

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructureImpl.java
```diff
@@ -427,6 +422,31 @@ public void stop() {
         deleteLocation();
     }
 
+    @Override
+    public void postStart(Collection<? extends Location> locations) {
+        super.postStart(locations);
+
+
+        if(config().get(DOCKER_SHOULD_START_REGISTRY)){
```
grkvlt: Formatting, including squashing multiple blank lines to single.

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
```diff
@@ -0,0 +1,110 @@
...
+                .poll(new HttpPollConfig<Boolean>(Attributes.SERVICE_UP)
+                        .onSuccess(Functions.constant(true))
+                        .onFailureOrException(Functions.constant(false)))
+                .poll(new HttpPollConfig<List<String>>(DOCKER_REGISTRY_CATALOG)
+                        .suburl("/_catalog")
+                        .onSuccess(Functionals.chain(
+                                HttpValueFunctions.jsonContents(),
+                                JsonFunctions.walk("repositories"),
+                                JsonFunctions.forEach(JsonFunctions.cast(String.class))))
+                        .onFailureOrException(Functions.constant(Arrays.asList(new String[0]))))
```
grkvlt: Maybe `Lists.<String>newArrayList()` instead of creating the empty `String[0]` array?

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
```diff
@@ -0,0 +1,110 @@
...
+        DockerHost host = (DockerHost) config().get(DOCKER_HOST);
+
+        config().set(DockerAttributes.DOCKER_PORT_BINDINGS, MutableMap.of(config().get(DOCKER_REGISTRY_PORT), 5000));
+
+        String installDir = host.sensors().get(SoftwareProcess.INSTALL_DIR);
+        config().set(DockerAttributes.DOCKER_HOST_VOLUME_MAPPING, MutableMap.of(Os.mergePaths(installDir, "certs"), "/certs"));
+
+        SshMachineLocation sshMachine = host.getDynamicLocation().getMachine();
+        String sshMachineInstallDir = getSSHMachineInstallDir();
+        sshMachine.installTo("classpath://brooklyn/entity/container/docker/registry/docker-registry-create-certs.sh", Os.mergePaths(sshMachineInstallDir, CREATE_CERTS_SCRIPT_NAME));
```
grkvlt: Put in a `public static final` constant at top of class

docker/src/main/java/brooklyn/entity/container/docker/repository/DockerRepositoryDriver.java
```diff
@@ -0,0 +1,6 @@
+package brooklyn.entity.container.docker.repository;
+
+import brooklyn.entity.container.docker.application.VanillaDockerApplicationDriver;
+
+public interface DockerRepositoryDriver extends VanillaDockerApplicationDriver {
```
grkvlt: Can we actually just reference the `VanillaDockerApplicationDriver` from our `DockerRepository` entity directly, without the need to create these subclasses?

patches/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
Graeme-Miller: done

Graeme-Miller:
Andrew, I have fixed everything as requested and added to the BOM files.
Can you please advise on those changes?
Also, you have requested examples written. Where do I put these examples?

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
```diff
@@ -0,0 +1,110 @@
...
+        DockerHost host = (DockerHost) config().get(DOCKER_HOST);
+
+        config().set(DockerAttributes.DOCKER_PORT_BINDINGS, MutableMap.of(config().get(DOCKER_REGISTRY_PORT), 5000));
+
+        String installDir = host.sensors().get(SoftwareProcess.INSTALL_DIR);
+        config().set(DockerAttributes.DOCKER_HOST_VOLUME_MAPPING, MutableMap.of(Os.mergePaths(installDir, "certs"), "/certs"));
+
+        SshMachineLocation sshMachine = host.getDynamicLocation().getMachine();
+        String sshMachineInstallDir = getSSHMachineInstallDir();
+        sshMachine.installTo(SCRIPT_LOCATION, Os.mergePaths(sshMachineInstallDir, CREATE_CERTS_SCRIPT_NAME));
```
grkvlt:
```Java
installTo(MutableMap.of(SshTool.PROP_PERMISSIONS, "0755"), SCRIPT_LOCATION, Os.mergePaths(sshMachineInstallDir, CREATE_CERTS_SCRIPT_NAME));
```

This should let you just call the script with a single command below.

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
```diff
@@ -0,0 +1,110 @@
...
+public class DockerRegistryImpl extends VanillaDockerApplicationImpl implements DockerRegistry {
+
+    private transient HttpFeed httpFeed;
+    private final String CREATE_CERTS_SCRIPT_NAME = "docker-registry-create-certs.sh";
+    private final String SCRIPT_LOCATION = "classpath://brooklyn/entity/container/docker/registry/docker-registry-create-certs.sh";
+
+    private String getSSHMachineInstallDir() {
+        return config().get(DockerRegistry.DOCKER_HOST).sensors().get(SoftwareProcess.INSTALL_DIR);
+    }
+
```
grkvlt: Add a `getIconUrl()` method that returns `classpath://docker-registry-logo.png`

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistry.java
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.brooklyn.core.sensor.Sensors;
+import org.apache.brooklyn.util.core.flags.SetFromFlag;
+
+import brooklyn.entity.container.DockerAttributes;
+import brooklyn.entity.container.docker.DockerContainer;
+import brooklyn.entity.container.docker.DockerHost;
+import brooklyn.entity.container.docker.application.VanillaDockerApplication;
+
+
+@ImplementedBy(DockerRegistryImpl.class)
```
grkvlt: Add the `@Catalog` annotations here

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructure.java
```diff
@@ -133,6 +133,11 @@
             new TypeToken<Map<String, Object>>() { },
             "docker.dockerfile.substitutions", "Dockerfile template substitutions", MutableMap.<String, Object>of());
 
+    @SetFromFlag("shouldStartRegistry")
```
grkvlt: Make this a catalog entry property, so we can see it when we start things from the catalog.

clocker.bom
```diff
@@ -49,6 +49,18 @@ brooklyn.catalog:
           $brooklyn:entitySpec:
             type: docker-host
 
+  - id: docker-infrastructure-with-registry
```
grkvlt: Should have a `docker-registry` catalog entry` and then add the config to a new template for a `docker-cloud-with-registry` as an example.

dist/src/main/assembly/conf/brooklyn/default.catalog.bom
```diff
@@ -193,26 +205,6 @@ brooklyn.catalog:
                     brooklyn.config:
                       docker.container.nameFormat: "weave-%2$02x"
 
-  - id: image-microservice
```
grkvlt: Why are these deleted? Should probably be replaced.

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistry.java
```diff
@@ -0,0 +1,46 @@
...
+public interface DockerRegistry extends VanillaDockerApplication {
+
+    @SetFromFlag("imageName")
```
grkvlt: Remove the `@SetFromFlag` for config that should not be overwritten

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistry.java
```diff
@@ -0,0 +1,46 @@
...
+    ConfigKey<String> IMAGE_NAME = ConfigKeys.newConfigKeyWithDefault(DockerAttributes.DOCKER_IMAGE_NAME.getConfigKey(), "registry");
+
+    @SetFromFlag("imageTag")
+    ConfigKey<String> IMAGE_TAG = ConfigKeys.newConfigKeyWithDefault(DockerAttributes.DOCKER_IMAGE_TAG.getConfigKey(), "2");
+
+    AttributeSensorAndConfigKey<Integer, Integer> DOCKER_REGISTRY_PORT = DockerHost.DOCKER_REGISTRY_PORT;
```
grkvlt: Add `@SetFromFlag` and `@CatalogConfig` here

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
Graeme-Miller: done

docker/src/main/java/brooklyn/entity/container/docker/registry/DockerRegistryImpl.java
Graeme-Miller: agreed that it was not worth re-writing at this stage

grkvlt:
Thanks @Graeme-Miller this looks good to merge into master

PR #178, opened 2015-10-14
Move away from the deprecated ServiceTypeResolver

neykov:
The changes in https://github.com/apache/incubator-brooklyn/pull/955 deprecate `ServiceTypeResolver` in favour of ServiceSpecResolver, BrooklynServiceTypeResolver is replaced by AbstractServiceSpecResolver. Update the code correspondingly.

The existing implementation will continue to work after the brooklyn PR is merged - there's a fallback for the deprecated interface.


docker/src/main/java/brooklyn/entity/container/docker/application/DockerServiceTypeResolver.java
```diff
@@ -72,16 +57,13 @@ public String getBrooklynType(String serviceType) {
         String imageName = Iterables.get(parts, 0);
         String imageTag = Iterables.get(parts, 1, "latest");
         log.debug("Creating Docker service entity with image {} and tag {}", imageName, imageTag);
+
+        EntitySpec<VanillaDockerApplication> spec = EntitySpec.create(VanillaDockerApplication.class);
         spec.configure(DockerAttributes.DOCKER_IMAGE_NAME, imageName);
         if (parts.size() == 2) {
             spec.configure(DockerAttributes.DOCKER_IMAGE_TAG, imageTag);
         }
-        if (resolver.getAttrs().containsKey("id")) {
```
grkvlt: @neykov How can the `ServiceSpecResolver` get information about the fields a service is configured with? I want to look up the `id: xxx` configuration here, is this still possible?

grkvlt:
This will only be required with Brooklyn versions _0.9.0-SNAPSHOT_ and above, so at the moment Clocker is not affected. Thanks for the heads up @neykov though. Will this mean things like `CAMP_PLAN_ID` are also deprecated?

docker/src/main/java/brooklyn/entity/container/docker/application/DockerServiceTypeResolver.java
neykov: Leaving out the attributes makes the resolvers independent of the parser using them. It's up to the parser to decide how to configure the EntitySpec based on the plan. We can still add them if there's a good reason, but in this particular case (and all other usages) it's not needed. The logic already in place in [`DockerUtils.getContainerName`](https://github.com/brooklyncentral/clocker/blob/master/docker/src/main/java/brooklyn/entity/container/DockerUtils.java#L169) will still populate `DOCKER_CONTAINER_NAME` from `BrooklynCampConstants.PLAN_ID` (which is `id` attribute) so the existing behaviour is preserved.

neykov:
> Will this mean things like CAMP_PLAN_ID are also deprecated?

No, it's set as before, what changes is that the resolver doesn't have access to the attributes. Turns out it wasn't really needed (see above comment for details).

docker/src/main/java/brooklyn/entity/container/docker/application/DockerServiceTypeResolver.java
grkvlt: Thanks @neykov, I agree it's not specifically needed here, but I was maybe wanting to do something more with this for Mesos and Marathon. I'll leave it for the moment, though.

neykov:
The Brooklyn PR is merged now, updated code to latest changes.

docker/src/main/java/brooklyn/entity/container/docker/application/DockerServiceTypeResolver.java
neykov: The PR was merged quickly so the TOSCA folks can use it, but we can still change it if needed.

grkvlt:
Now included in #201

PR #184, opened 2015-11-10
Clocker UI customization

grkvlt:
Copy of the Brooklyn `usage/jsgui` webapp with Clocker styling


grkvlt:
clocker-jsgui-1
clocker-jsgui-2

PR #185, opened 2015-11-10
Integration for the Apache Mesos Marathon framework

grkvlt:


grkvlt:
This supports `VanillaDockerApplication` types deployed as `docker:image:tag` which will create a `MarathonTask` in a connected Mesos deployment. For example, the following blueprint will create a Redis server named _test_ from a Docker image.

```YAML
location: my-mesos-cluster
services:
- type: docker:redis:3
  id: test
  openPorts:
  - 6379
  provisioningProperties:
    minRam: 1024
```

marathon-redis-deploy

@ahgittin @neykov `SoftwareProcess` is still a bit broken, but that is covered in another pull request I am working on. This is being merged to allow further testing while those remaining issues with SSH are sorted out.

PR #192, opened 2015-11-13
Mesos authentication support

neykov:
Plus doesn't fail on self signed certificates.


grkvlt:
@neykov thanks for this, my test environment doesn't have authentication - is this ready to merge?

PR #198, opened 2015-11-26
Support Calico in Mesos

grkvlt:
Add a new `CalicoModule` SDN provider to support Calico in Mesos and Marathon


grkvlt:
```YAML
id: calico-mesos-cluster
name: "Calico Mesos Cluster"
description: |
  Mesos cluster with Marathon and Calico enabled.

services:
- type: mesos-cluster
  brooklyn.config:
    entity.dynamicLocation.name: "my-mesos-cluster"
    mesos.url: "http://mesos:5050/"
    mesos.slave.accessible: true
    mesos.slave.ssh.user: "root"
    mesos.slave.ssh.privateKey.file: ".ssh/id_rsa"
    sdn.enable: true
    sdn.provider.spec:
      $brooklyn:entitySpec:
        type: calico-module
        brooklyn.config:
          etcd.cluster.url: "mesos:4001"
          sdn.network.cidr: 50.0.0.0/16
          sdn.network.size: 24
```

neykov:
Got a `sudo: sorry, you must have a tty to run sudo`, because it's attaching to an existing machine, not configured by jclouds.

grkvlt:
@neykov You need the following lines in `brooklyn.properties` to make Brooklyn work with the CentOS VMs that MANTL provisions:
```
## make SSH work with sudo on CentOS ##
brooklyn.ssh.config.tool.class = org.apache.brooklyn.util.core.internal.ssh.cli.SshCliTool
brooklyn.ssh.config.sshFlags = -tt -vvv
```

neykov:
After manually commenting out `requiretty` in the ssh config it's working great with the default tool.
Could add a call to `SshTasks.dontRequireTtyForSudo` so it doesn't require custom config?

grkvlt:
@neykov Yes, could you do that and send me the patch and I'll add it here

grkvlt:
Note that this only works with the MANTL provisioned Mesos clusters, as this uses the same Docker 1.7.x and Calico 0.4.x that Clocker uses. Future revisions should support the libnetwork Calico with Docker 1.9.x onwards, once this is also available in the MANTL deployments.

grkvlt:
@neykov There's also `SshTool.PROP_ALLOCATE_PTY`which I'm going to test

grkvlt:
This should be ready to merge now. Screenshot shows Brooklyn entity and Docker image running in MANTL.

mesos-empty-software-process

PR #201, opened 2015-11-26
Build against Brooklyn 0.9.0-SNAPSHOT

johnmccabe:
Requires brooklyncentral/advanced-networking/pull/85.

Tested docker cloud with calico in Softlayer, deployed Node catalog app, Tomcat and redis/dnmonster/identidock examples from getting started section on http://clocker.io.

*Note* the cast on line 617 in DockerContainerImpl needs to be properly fixed.


johnmccabe:
Hopefully resolves #191

grkvlt:
Thanks @johnmccabe we'll merge this once a 1.1.0 release has been made

grkvlt:
@johnmccabe suggest you cherry-pick commit 5eaf90ff267a702c9d9da6e322d393fd953e9c1e from #178 as well

docker/src/main/java/brooklyn/entity/container/docker/DockerContainerImpl.java
```diff
@@ -614,13 +614,15 @@ public DockerContainerLocation createLocation(Map flags) {
             }
 
             // Create our wrapper location around the container
-            LocationSpec<DockerContainerLocation> spec = LocationSpec.create(DockerContainerLocation.class)
+            LocationSpec<DockerContainerLocation> spec = (LocationSpec<DockerContainerLocation>) LocationSpec.create(DockerContainerLocation.class)
```
Graeme-Miller: why is this required?

Graeme-Miller: This compiles, might be neater than casting:
```java
LocationSpec<DockerContainerLocation> spec = LocationSpec.create(DockerContainerLocation.class);
            spec.parent(host)
            .configure(flags)
...
```

but still not sure what has changed to make this required

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructure.java
```diff
@@ -86,12 +86,15 @@
...
-    ConfigKey<EntitySpec> SDN_PROVIDER_SPEC = ConfigKeys.newConfigKey(EntitySpec.class, "sdn.provider.spec", "SDN provider entity specification");
+    ConfigKey<EntitySpec<?>> SDN_PROVIDER_SPEC = ConfigKeys.newConfigKey(
+            new TypeToken<EntitySpec<?>>() {}, "sdn.provider.spec", "SDN provider entity specification");
 
+    @SuppressWarnings("serial")
     @SetFromFlag("hostSpec")
-    AttributeSensorAndConfigKey<EntitySpec, EntitySpec> DOCKER_HOST_SPEC = ConfigKeys.newSensorAndConfigKey(
-            EntitySpec.class, "docker.host.spec", "Specification to use when creating child Docker Hosts",
+    AttributeSensorAndConfigKey<EntitySpec<?>, EntitySpec<?>> DOCKER_HOST_SPEC = ConfigKeys.newSensorAndConfigKey(
+            new TypeToken<EntitySpec<?>>() {}, "docker.host.spec", "Specification to use when creating child Docker Hosts",
```
Graeme-Miller: as above, why is this change required?

docker/src/main/java/brooklyn/entity/container/docker/DockerContainerImpl.java
johnmccabe: Changing as suggested, didn't come up with a better way of dealing with the type erasure.

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructure.java
grkvlt: Yes, I'd prefer to leave this (and similar instances) _without_ the generic types rather than explicitly saying `Blah<?>` everywhere, and having to use `TypeToken<Blah<?>>` instead of just `Blah.class` since it adds no extra information.

johnmccabe: @grkvlt I've stripped out the generics that I'd introduced from the interfaces, but have not figured out how to avoid the cast to `EntitySpec<?>` in [DockerInfrastructureImpl.java#L185](https://github.com/johnmccabe/clocker/blob/0.9.0-update/docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructureImpl.java#L185]) and [MesosClusterImpl.java#L149](https://github.com/johnmccabe/clocker/blob/0.9.0-update/mesos/src/main/java/brooklyn/entity/mesos/MesosClusterImpl.java#L149).

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructureImpl.java
```diff
@@ -182,7 +182,7 @@ public String apply(@Nullable Entity input) {
                 .displayName("Docker Applications"));
 
         if (config().get(SDN_ENABLE) && config().get(SDN_PROVIDER_SPEC) != null) {
-            Entity sdn = addChild(EntitySpec.create(config().get(SDN_PROVIDER_SPEC))
+            Entity sdn = addChild(EntitySpec.create((EntitySpec<?>)config().get(SDN_PROVIDER_SPEC))
```
grkvlt: We don't use the results of this cast for anything, and in fact the return type of `addChild(EntitySpec)` is just `Entity` so it should be removed

docker/src/main/java/brooklyn/entity/container/docker/application/DockerEntitySpecResolver.java
```diff
@@ -19,51 +19,37 @@
...
-import org.apache.brooklyn.camp.brooklyn.spi.creation.BrooklynComponentTemplateResolver;
-import org.apache.brooklyn.camp.brooklyn.spi.creation.service.BrooklynServiceTypeResolver;
-import org.apache.brooklyn.camp.brooklyn.spi.creation.service.ServiceTypeResolver;
-import org.apache.brooklyn.camp.spi.PlatformComponentTemplate;
-import org.apache.brooklyn.util.text.Strings;
-
 import brooklyn.entity.container.DockerAttributes;
 
 /**
  * This converts {@link PlatformComponentTemplate} instances whose type is prefixed {@code docker:}
```
grkvlt: The Javadoc comments should be revised in line with the code changes

docker/src/main/java/brooklyn/entity/container/docker/application/DockerEntitySpecResolver.java
```diff
@@ -71,16 +57,12 @@ public String getBrooklynType(String serviceType) {
...
         spec.configure(DockerAttributes.DOCKER_IMAGE_NAME, imageName);
         if (parts.size() == 2) {
             spec.configure(DockerAttributes.DOCKER_IMAGE_TAG, imageTag);
         }
-        if (resolver.getAttrs().containsKey("id")) {
-            String containerName = (String) resolver.getAttrs().getStringKey("id");
-            spec.configure(DockerAttributes.DOCKER_CONTAINER_NAME, containerName);
-        }
-
-        super.decorateSpec(resolver, spec);
```
grkvlt: No call to `super` needed?

mesos/src/main/java/brooklyn/entity/mesos/MesosClusterImpl.java
```diff
@@ -146,7 +146,7 @@ public String apply(@Nullable Entity input) {
         }
 
         if (config().get(SDN_ENABLE) && config().get(SDN_PROVIDER_SPEC) != null) {
-            Entity sdn = addChild(EntitySpec.create(config().get(SDN_PROVIDER_SPEC))
+            Entity sdn = addChild(EntitySpec.create((EntitySpec<?>)config().get(SDN_PROVIDER_SPEC))
```
grkvlt: See previous comment; cast not required here

grkvlt:
https://github.com/johnmccabe/clocker/blob/0.9.0-update/docker/src/main/resources/META-INF/services/org.apache.brooklyn.core.resolve.ServiceSpecResolver

This file needs changed to point at the new `DockerEntitySpecResolver` class

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructureImpl.java
johnmccabe: Problem here is that without the cast what gets passed to the child is `org.apache.brooklyn.api.internal.AbstractBrooklynObjectSpec`. It's the `configure` method that's causing the problem but I don't currently see how/where to resolve it without the cast, previously `configure` was implemented in the `EntitySpec` and we didn't encounter this problem, but it now resides in `AbstractBrooklynObjectSpec` and we're losing the type along the way. I'm going to try to replicate what I'm seeing to try to understand it.

docker/src/main/java/brooklyn/entity/container/docker/application/DockerEntitySpecResolver.java
johnmccabe: This is part of the cherry picked commit https://github.com/brooklyncentral/clocker/commit/5eaf90ff267a702c9d9da6e322d393fd953e9c1e from https://github.com/brooklyncentral/clocker/pull/178, I just merged and deployed a few blueprints to check, I didn't dig into the specifics. I'll have a closer look.

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructureImpl.java
Graeme-Miller: Similar trick to last time maybe?
```YAML
EntitySpec entitySpec = config().get(SDN_PROVIDER_SPEC);
entitySpec.configure(DockerAttributes.DOCKER_INFRASTRUCTURE, this);
Entity sdn = addChild(EntitySpec.create(entitySpec));
```

johnmccabe: Thanks @Graeme-Miller that works.

docker/src/main/java/brooklyn/entity/container/docker/application/DockerEntitySpecResolver.java
neykov: That's an abstract method override, no super implementation.

neykov:
Also worth looking through the `patches` project to remove already integrated changes.

Graeme-Miller:
@neykov I have been looking into Clocker patches recently. Most of the patches are JClouds, and we will need a new release of JClouds to get them.

There is one patch that we should be able to get rid of, and that is JcloudsLocationSecurityGroupCustomizer (despite the name it is a brooklyn class).

PR #206, opened 2015-11-30
Fix container names

Graeme-Miller:
Recreated this PR, as it wouldn't let me re-open the original.
Whilst we now have config with a default to make the scenario unlikely, the code will still fall over when "getContainerName(entity)" returns a value but "getDockerContainerName()" does not. I think the code should only throw an exception when both are null, hence this change.


grkvlt:
@Graeme-Miller Under what circumastance is `getDockerContainerName()` null? Can we fix that instead?

Graeme-Miller:
@grkvlt whenever "docker.container.name" and "docker.container.nameFormat" are not specified on the docker container. It is entirely possible, although unlikely, that a user will configure this incorrectly.

PR #209, opened 2015-12-01
Extra public key config

neykov:
Based on https://github.com/brooklyncentral/clocker/pull/198, see last commit.


neykov:
Rebased on master.

grkvlt:
@neykov we should also support passing in a public key file URL, I think, or an alternative `authorized_keys` file in its entirity.

PR #211, opened 2015-12-03
Added volume and entrypoint config to microservices

grkvlt:
- Added extra config to `MicroserviceImage` and `MicroserviceDockerfile` for volumes.
- Includes support for setting the entrypoint for a container.
- Adds `portBindings` configuration keys

### DIFF
--- a/docker/src/main/java/brooklyn/entity/container/docker/application/VanillaDockerApplication.java
+++ b/docker/src/main/java/brooklyn/entity/container/docker/application/VanillaDockerApplication.java
@@ -54,6 +54,9 @@ public interface VanillaDockerApplication extends VanillaSoftwareProcess {
     @SetFromFlag("imageTag")
     ConfigKey<String> IMAGE_TAG = DockerAttributes.DOCKER_IMAGE_TAG.getConfigKey();
 
+    @SetFromFlag("entrypoint")
+    ConfigKey<List<String>> IMAGE_ENTRYPOINT = DockerAttributes.DOCKER_IMAGE_ENTRYPOINT.getConfigKey();
+
     @SetFromFlag("useSsh")
     ConfigKey<Boolean> DOCKER_USE_SSH = ConfigKeys.newConfigKeyWithDefault(DockerAttributes.DOCKER_USE_SSH, Boolean.FALSE);
 
--- a/mesos/src/main/java/brooklyn/entity/mesos/task/marathon/MarathonTaskImpl.java
+++ b/mesos/src/main/java/brooklyn/entity/mesos/task/marathon/MarathonTaskImpl.java
@@ -493,7 +493,7 @@ public class MarathonTaskImpl extends MesosTaskImpl implements MarathonTask {
             // Docker command or args
             String command = config().get(COMMAND);
             builder.putIfNotNull("command", command);
-            List<String> args = config().get(ARGS);
+            List<String> args = MutableList.copyOf(config().get(ARGS));
             builder.putIfNotNull("args", args);
         } else {
             // OS name for image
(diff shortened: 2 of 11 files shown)


grkvlt:
@johnmccabe @Graeme-Miller This should be useful for microservice blueprints. Can you test the new config options?

docker/src/main/java/brooklyn/entity/container/docker/application/VanillaDockerApplication.java
```diff
@@ -54,6 +54,9 @@
     @SetFromFlag("imageTag")
     ConfigKey<String> IMAGE_TAG = DockerAttributes.DOCKER_IMAGE_TAG.getConfigKey();
 
+    @SetFromFlag("entrypoint")
+    ConfigKey<List<String>> IMAGE_ENTRYPOINT = DockerAttributes.DOCKER_IMAGE_ENTRYPOINT.getConfigKey();
```
Graeme-Miller: should this be called DOCKER_IMAGE_ENTRYPOINT for consistency?

mesos/src/main/java/brooklyn/entity/mesos/task/marathon/MarathonTaskImpl.java
```diff
@@ -493,7 +493,7 @@ public void stop() {
             // Docker command or args
             String command = config().get(COMMAND);
             builder.putIfNotNull("command", command);
-            List<String> args = config().get(ARGS);
+            List<String> args = MutableList.copyOf(config().get(ARGS));
```
Graeme-Miller: is it not possible to have other args for a marathon task other than an image entrypoint?

Graeme-Miller:
I can't see where DOCKER_IMAGE_ENTRYPOINT is actually being used when creating a docker image or starting a docker container

grkvlt:
@Graeme-Miller see the `DockerContainerImpl` and `MarathonTaskImpl` changes. Note that because Marathon OS just launching Docker images, the `args` parameter is exactly equivalent to the Docker entry point for the container.

docker/src/main/java/brooklyn/entity/container/docker/application/VanillaDockerApplication.java
grkvlt: I am trying to be consistent with the other configuration keys here, hence the aliasing of the field name

mesos/src/main/java/brooklyn/entity/mesos/task/marathon/MarathonTaskImpl.java
grkvlt: No. See my other comment

johnmccabe:
@grkvlt trying with this - https://gist.github.com/johnmccabe/e67a5da18d5629aad6e0, can't get the volumes to work as expected yet tho

grkvlt:
@johnmccabe I will merge this now, and we can work out the issue with the underlying volume support in another PR

docker/src/main/java/brooklyn/entity/container/docker/application/VanillaDockerApplication.java
Graeme-Miller: would it make sense to rename the other config in this file as well then, so that everything matches there corresponding variables in DockerAttributes?

Graeme-Miller:
@grkvlt OK, that makes sense for Marathon. However, I can't see how we are using DOCKER_IMAGE_ENTRYPOINT for DockerContainers. The change in "DockerContainerImpl" sets this as a sensor, how does that impact the entry point when running the container? I would have expected some changes to DockerHostLocation.obtain, unless I have misunderstood the purpose of this config?

grkvlt:
@Graeme-Miller it's set in `DockerTemplateOptions` here https://github.com/brooklyncentral/clocker/blob/master/docker/src/main/java/brooklyn/entity/container/docker/DockerContainerImpl.java#L425

PR #218, opened 2015-12-08
Update Docker installation 

Graeme-Miller:
Update Docker installation to use new Docker apt-get repositories. This will allow us to use docker 1.8 and 1.9

### DIFF
--- a/docker/src/main/java/brooklyn/entity/container/docker/DockerHostSshDriver.java
+++ b/docker/src/main/java/brooklyn/entity/container/docker/DockerHostSshDriver.java
@@ -289,13 +289,38 @@ public class DockerHostSshDriver extends AbstractSoftwareProcessSshDriver implem
     }
 
     private String installDockerOnUbuntu() {
-        String version = getVersion();
-        log.debug("Installing Docker version {} on Ubuntu", version);
+        String dockerVersion = getVersion();
+        String ubuntuVersion = getMachine().getMachineDetails().getOsDetails().getVersion();
+
+        String dockerRepoName;
+        String repositoryVersionName;
+        switch (ubuntuVersion) {
+            case "12.04":
+                dockerRepoName = "ubuntu-precise";
+                repositoryVersionName = dockerVersion + "-0~precise";
+                break;
+            case "14.04":
+                dockerRepoName = "ubuntu-trusty";
+                repositoryVersionName = dockerVersion + "-0~trusty";
+                break;
+            case "15.04":
+                dockerRepoName = "ubuntu-vivid";
+                repositoryVersionName = dockerVersion + "-0~vivid";
+                break;
+            case "15.10":
+                dockerRepoName = "ubuntu-wily";
+                repositoryVersionName = dockerVersion + "-0~wily";
+                break;
+            default:
+                throw new IllegalArgumentException("No docker repo found for ubuntu version: " + ubuntuVersion);
+        }
+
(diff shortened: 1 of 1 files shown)


docker/src/main/java/brooklyn/entity/container/docker/DockerHostSshDriver.java
```diff
@@ -289,13 +289,25 @@ public String getVersion() {
...
 
     private String installDockerOnUbuntu() {
-        String version = getVersion();
-        log.debug("Installing Docker version {} on Ubuntu", version);
+        final String dockerVersion = getVersion();
+        final String ubuntuVersion = getMachine().getMachineDetails().getOsDetails().getVersion();
+
+        final String dockerRepoName;
+        final String repositoryVersionName;
+        switch(ubuntuVersion){
```
grkvlt: Add spaces: `switch (ubuntuVersion) {`

docker/src/main/java/brooklyn/entity/container/docker/DockerHostSshDriver.java
```diff
@@ -289,13 +289,25 @@ public String getVersion() {
...
+            case "12.04": dockerRepoName = "ubuntu-precise"; repositoryVersionName = dockerVersion+ "-0~precise"; break;
```
grkvlt: split all of these onto multiple lines:
```Java
case "12.04":
    dockerRepoName = "ubuntu-precise"; repositoryVersionName = dockerVersion+ "-0~precise";
    break;
```

docker/src/main/java/brooklyn/entity/container/docker/DockerHostSshDriver.java
```diff
@@ -289,13 +289,25 @@ public String getVersion() {
     }
 
     private String installDockerOnUbuntu() {
-        String version = getVersion();
-        log.debug("Installing Docker version {} on Ubuntu", version);
+        final String dockerVersion = getVersion();
```
grkvlt: Not sure why any of these are `final` so suggest removing it from all

docker/src/main/java/brooklyn/entity/container/docker/DockerHostSshDriver.java
```diff
@@ -289,13 +289,25 @@ public String getVersion() {
...
+        final String ubuntuVersion = getMachine().getMachineDetails().getOsDetails().getVersion();
+
```
grkvlt: Remove this blank line but add one before the `switch` statement

docker/src/main/java/brooklyn/entity/container/docker/DockerHostSshDriver.java
Graeme-Miller: It's good practice
"In general, declare all variables final if they are not modified after being initialized or set to some value."
https://docs.oracle.com/cd/E19159-01/819-3681/abebf/index.html

docker/src/main/java/brooklyn/entity/container/docker/DockerHostSshDriver.java
Graeme-Miller: According to Oracle's programming style, this is correct:
"One blank line should always be used in the following circumstances:
Between the local variables in a method and its first statement
..
Between logical sections inside a method to improve readability"
http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-141388.html

PR #225, opened 2015-12-18
Ensure unique cointainer names if plan ID is set

neykov:
Plan IDs are not guaranteed to be unique even in a single application. This becomes a problem especially when using catalog items - can reference them only once per docker host.
Ensure unique container names when generated from the plan id.

### DIFF
--- a/docker/src/main/java/brooklyn/entity/container/DockerUtils.java
+++ b/docker/src/main/java/brooklyn/entity/container/DockerUtils.java
@@ -224,8 +224,18 @@ public class DockerUtils {
     public static Optional<String> getContainerName(Entity target) {
         return Optional.fromNullable(target.sensors().get(DockerContainer.DOCKER_CONTAINER_NAME))
                 .or(Optional.fromNullable(target.config().get(DockerContainer.DOCKER_CONTAINER_NAME)))
-                .or(Optional.fromNullable(target.config().get(BrooklynCampConstants.PLAN_ID)))
+                .or(Optional.fromNullable(getContainerNameFromPlan(target)))
                 .transform(DockerUtils.ALLOWED);
+    }
+
+    private static String getContainerNameFromPlan(Entity target) {
+        String planId = target.config().get(BrooklynCampConstants.PLAN_ID);
+        if (planId != null) {
+            //Plan IDs are not unique even in a single application
+            return planId + "_" + target.getId();
+        } else {
+            return null;
+        }
     }
 
     /* Generate the address to use to talk to another target entity. */
--- a/mesos/src/main/java/brooklyn/location/mesos/framework/marathon/MarathonLocation.java
+++ b/mesos/src/main/java/brooklyn/location/mesos/framework/marathon/MarathonLocation.java
@@ -161,6 +161,7 @@ public class MarathonLocation extends MesosFrameworkLocation implements MachineP
         String planId = entity.config().get(BrooklynCampConstants.PLAN_ID);
         String name = null;
         if (planId != null) {
+            //Plan IDs are not unique even in a single application
             name = planId + "/" + entity.getId();
         } else {
             name = entity.getId();


Graeme-Miller:
No opposition to this change, as it increases our resilience when we need to fallback to plan id. However, I would have thought that the DockerContainer.DOCKER_CONTAINER_NAME_FORMAT config would have taken care of unique container IDs. [DockerContainerImpl line 115](https://github.com/brooklyncentral/clocker/blob/master/docker/src/main/java/brooklyn/entity/container/docker/DockerContainerImpl.java#L115) generates a name from the DOCKER_CONTAINER_NAME_FORMAT config.

docker/src/main/java/brooklyn/entity/container/DockerUtils.java
```diff
@@ -224,10 +224,20 @@ public static String imageName(Entity entity, String dockerfile) {
...
                 .transform(DockerUtils.ALLOWED);
     }
 
+    private static String getContainerNameFromPlan(Entity target) {
+        String planId = target.config().get(BrooklynCampConstants.PLAN_ID);
+        if (planId != null) {
+            //Plan IDs are not unique even in a single application
+            return planId + "_" + target.getId();
+        } else {
+            return null;
```
ahgittin: why not

    return target.getId();

mesos/src/main/java/brooklyn/location/mesos/framework/marathon/MarathonLocation.java
```diff
@@ -161,6 +161,7 @@ private String getTaskName(Entity entity) {
         String planId = entity.config().get(BrooklynCampConstants.PLAN_ID);
```
ahgittin: and could reuse

    DockerUtils.getEntitiyIdentifierForContainerName(Entity target, String separator)

based on `getContainerNameFromPlan`

neykov:
@Graeme-Miller Having a PLAN_ID value overrides the auto-generated name. The auto-increment is not appended to the override.

docker/src/main/java/brooklyn/entity/container/DockerUtils.java
neykov: It would prevent the existing logic from working as expected. If no `PLAN_ID`, then it would use an auto-generated name (see the link in Graeme's post).

mesos/src/main/java/brooklyn/location/mesos/framework/marathon/MarathonLocation.java
neykov: Here we don't have auto-generated value, so the logic is a bit different, Makes sense to reuse the logic, but at a higher level.

Graeme-Miller:
@neykov could you point me to the location in the code?
My understanding was that [DockerUtils.getContainerName](https://github.com/neykov/clocker/blob/unique-container-name/docker/src/main/java/brooklyn/entity/container/DockerUtils.java#L225) checked the DOCKER_CONTAINER_NAME config first. Isn't this set by the init method I linked to earlier?

neykov:
@Graeme-Miller The init method sets the name on the sensor of the **container** entity. On the other hand `DockerUtils.getContainerName` is called on the entity being deployed in the container location, which doesn't have a sensor set.

grkvlt:
I think the order should be:

1. DOCKER_CONTAINER_NAME config key (from entity being deployed)
2. PLAN_ID and entity id (from entity being deployed)
3. DOCKER_CONTAINER_NAME_FORMAT (from the host)

This fixes the issue of the plan id not being unique. I would like to be able to use the PLAN_ID, since many docker images want to refer to a hostname that is the same as the container name. But, this could be different, since when adding a line to /etc/hosts it can be added as an alias. This would mean some further changes to the code for adding container links to allow the changes here to work with e.g. the Identidock demo application.

The sensor value gets set on the DockerContainer entity and the entity being deployed, which we can do because it is the entity that causes the container to be created.

PR #227, opened 2016-01-12
re-add ipip and nat-outgoing to calico command

Graeme-Miller:


grkvlt:
Looks good. I will investigate changing the Mesos drivers for Calico as well.

PR #230, opened 2016-01-21
Modify the start() method to deal with latest Brooklyn semantics

grkvlt:
Changed all `#start(Collection)` methods to use `Locations#getLocationsCheckingAncestors(Collection, Entity)`


Graeme-Miller:
lgtm

docker/src/main/java/brooklyn/entity/container/docker/DockerContainerImpl.java
```diff
@@ -606,7 +607,12 @@ public void deleteLocation() {
     }
 
     @Override
-    public void start(Collection<? extends Location> locations) {
+    public void start(Collection<? extends Location> locs) {
+        if (locs != null && !locs.isEmpty()) {
```
ahgittin: this `if` check is no longer necessary - `addLocations` has been changed to accept null/empty and do no-op.

grkvlt: OK, will change.

Issue #231, opened 2016-01-21
Fix mapped.main.uri


Graeme-Miller:
Clocker UI currently displays the wrong URI on the Summary page for an entity.
It displays main.uri, but should display mapped.main.uri

grkvlt:
Should check for presence of mapped URI in the JS code for the summary pane.

grkvlt:
Fixed in https://github.com/brooklyncentral/clocker/commit/0ee83fd42581eb73c1255e48f6a06c2263b754e5

PR #233, opened 2016-02-11
Main URI Mapping fix

Graeme-Miller:
Add a list of sensors that are not to be mapped, they should just be translated instead. Added MAIN_URI to that list.

This will fix issue #231.


grkvlt:
LGTM, merging.

PR #236, opened 2016-02-21
Fix issue where Marathon ports are not mapped correctly

grkvlt:
The `mapped.docker.port.XXXX` sensors were not being populated, due to a delay in setting the location on entities.


johnmccabe:
fyi @neykov

PR #239, opened 2016-02-22
Added placement strategy for grouping containers by parent

grkvlt:


grkvlt:
Merging to make available for testing

PR #241, opened 2016-02-23
Karaf features for clocker

CMoH:
I had to stop at brooklyn-software-jclouds, which uses a version of guava incompatible with that used by jclouds (at least from a semantic versioning perspective). This problem is still unresolved in brooklyn-features, so it spilled here.

We can disable validation for the time being, but this will only help making the build succeed - but installing those features in karaf will still fail.


grkvlt:
@CMoH thanks, I'll have a look

grkvlt:
Closed as superseded by #246

PR #243, opened 2016-02-29
Removed DynamicGroup from framework entities and updated start logic

grkvlt:
Makes startup sequence simpler and prevents *ON_FIRE* state for container and task groups.


grkvlt:
Tested OK, merging

PR #250, opened 2016-03-11
Added automatic QA

Graeme-Miller:
Added Automatic QA. There is a new profile in the Clocker dist maven that will upload the dist to AWS, and then kickoff a test run using the QA framework.

I have also included travis configuration to do this automatically.

I have made a couple of changes to core Clocker code. Namely:
- Added some PLAN_IDs. This is to allow these components to be reference by test entities
- Removed 'my-docker-cloud' default location name. Need to add multiple infrastructures without setting the name for testing. Infrastructures will now be give a name with an auto incremented ID added

### DIFF
--- a/docker/src/main/java/brooklyn/entity/container/docker/DockerHostImpl.java
+++ b/docker/src/main/java/brooklyn/entity/container/docker/DockerHostImpl.java
@@ -64,6 +64,7 @@ import org.apache.brooklyn.api.location.PortRange;
 import org.apache.brooklyn.api.mgmt.LocationManager;
 import org.apache.brooklyn.api.mgmt.Task;
 import org.apache.brooklyn.api.policy.PolicySpec;
+import org.apache.brooklyn.camp.brooklyn.BrooklynCampConstants;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.render.RendererHints;
 import org.apache.brooklyn.core.effector.ssh.SshEffectorTasks;
@@ -176,6 +177,7 @@ public class DockerHostImpl extends MachineEntityImpl implements DockerHost {
         Group containers = addChild(EntitySpec.create(BasicGroup.class)
                 .configure(BasicGroup.RUNNING_QUORUM_CHECK, QuorumChecks.atLeastOneUnlessEmpty())
                 .configure(BasicGroup.UP_QUORUM_CHECK, QuorumChecks.atLeastOneUnlessEmpty())
+                .configure(BrooklynCampConstants.PLAN_ID, "docker-host-containers")
                 .displayName("Docker Containers"));
         if (config().get(DockerInfrastructure.HA_POLICY_ENABLE)) {
             containers.policies().add(PolicySpec.create(ServiceReplacer.class)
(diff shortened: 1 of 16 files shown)


docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructure.java
```diff
@@ -57,7 +58,7 @@
 
     @CatalogConfig(label = "Location Name", priority = 90)
     @SetFromFlag("locationName")
-    AttributeSensorAndConfigKey<String, String> LOCATION_NAME = ConfigKeys.newSensorAndConfigKeyWithDefault(LocationOwner.LOCATION_NAME, "my-docker-cloud");
```
grkvlt: We want the default in the wizard for the catalog entry, I think.

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructureImpl.java
```diff
@@ -101,6 +102,8 @@
 
     private transient Object mutex = new Object[0];
 
+    private AtomicInteger infrastructureID = new AtomicInteger();
```
grkvlt: I think this should be `static final` so that it is shared across all instances,?

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructureImpl.java
```diff
@@ -162,6 +165,7 @@ public void init() {
                 .configure(DynamicCluster.MEMBER_SPEC, dockerHostSpec)
                 .configure(DynamicCluster.RUNNING_QUORUM_CHECK, QuorumChecks.atLeastOneUnlessEmpty())
                 .configure(DynamicCluster.UP_QUORUM_CHECK, QuorumChecks.atLeastOneUnlessEmpty())
+                .configure(BrooklynCampConstants.PLAN_ID, "DockerHosts")
```
grkvlt: We use `CaseFormat.LOWER_HYPHEN` for plan IDs, so `docker-hosts` here.

docker/src/main/java/brooklyn/entity/container/docker/DockerHostImpl.java
```diff
@@ -176,6 +177,7 @@ public void init() {
         Group containers = addChild(EntitySpec.create(BasicGroup.class)
                 .configure(BasicGroup.RUNNING_QUORUM_CHECK, QuorumChecks.atLeastOneUnlessEmpty())
                 .configure(BasicGroup.UP_QUORUM_CHECK, QuorumChecks.atLeastOneUnlessEmpty())
+                .configure(BrooklynCampConstants.PLAN_ID, "DockerHostContainers")
```
grkvlt: `docker-host-containers`

docker/src/main/java/brooklyn/entity/container/docker/DockerInfrastructureImpl.java
```diff
@@ -101,6 +102,8 @@
 
     private transient Object mutex = new Object[0];
 
+    private static final AtomicInteger infrastructureID = new AtomicInteger();
```
grkvlt: Forgot, we probably also want to stuff this into an `AttributeSensor<AtomicInteger>` for persistence.

grkvlt:
Merged. Will make review changes in a separate commit.

andreaturli:
ok, in fact I'm not sure we should leave things like https://github.com/brooklyncentral/clocker/pull/250/files#diff-6162d9f90c325fae4c9c945b5ded55d4R213

grkvlt:
Good catch @andreaturli but it does require a login...

grkvlt:
https://github.com/brooklyncentral/clocker/commit/5a336ee4ae0b6da453c7f5ac5a09b3f20aadba38 fixes issue of `AtomicInteger` persistence by removing it :cat:

Issue #237, opened 2016-02-21
NPE rebinding to clocker persisted state - causes Brooklyn server restart to fail


aledsage:
When restarting Brooklyn with a couple of clocker+calico apps (as deployed from the default catalog)...

I got the `NullPointerException` shown below.

In brief, `DockerInfrastructureImpl.rebind()` calls its PROPERTIES_RELOAD_LISTENER, which `ReloadDockerLocation.reloaded()`. This tries to resolve the spec. However, the `DockerLocation` instances have not been fully rebound yet - so the call to `dockerLocation.owner()` returns null (even though it will eventually have its "owner" reset.

I suspect that if we could delay the call to the reload listener until later, then we wouldn't get this NPE.

Guarding against the NPE in `DockerResolver` is probably not a good idea - it would create an identical location, so we'd get a duplicate.

```
2016-02-21 16:50:54,434 ERROR o.a.b.l.BrooklynLauncher [main]: Subsystem for persistence had startup error (continuing with startup): org.apache.brooklyn.util.exceptions.PropagatedRuntimeException: org.apache.brooklyn.util.exceptions.PropagatedRuntimeException: Failure 
rebinding: problem rebinding entity j1M7T5XG (DockerInfrastructureImpl{id=j1M7T5XG}): Cannot instantiate location 'LocationDefinition{id='gzuSFMac', name='my-docker-cloud-vagrant', spec='docker:j1M7T5XG:(name="my-docker-cloud-vagrant")', config={provisioner=FixedListMac
hineProvisioningLocation{id=DtPfBrpb, name=FixedListMachineProvisioningLocation:DtPf}, strategies=[DockerAwarePlacementStrategy(MaxContainersPlacementStrategy@j8yVexHt), DockerAwarePlacementStrategy(BreadthFirstPlacementStrategy@ce2SXY0W)]}}' pointing at docker:j1M7T5XG
:(name="my-docker-cloud-vagrant"): java.lang.NullPointerException
org.apache.brooklyn.util.exceptions.PropagatedRuntimeException: 
        at org.apache.brooklyn.util.exceptions.Exceptions.propagate(Exceptions.java:108) ~[brooklyn-utils-common-0.9.0-20160125.1743-p1.jar:0.9.0-20160125.1743-p1]
[...]

PR #252, opened 2016-03-19
Various fixes (see individual commits).

aledsage:
Requires that https://github.com/apache/brooklyn-server/pull/59 be merged first.

Highlights include:

* Fixes location-definition persistence. See https://github.com/apache/brooklyn-server/pull/59 and https://github.com/brooklyncentral/clocker/issues/237
* For ssh'able containers, don't just rely on the root:password that the dockerfile contains. Instead use the standard `JcloudsLocation` mechanism to change the user and add to authorized_keys.
* By default, disable the automatic image creation at the end of the "install" phase of SoftwareProcess entities. This image creation approach is flawed for containers with sshd because (now that we allow the user/credential to be configured properly) the checkpointed image would include the container's custom username/credential. This makes logging into a new container problematic!
* The DockerInfrastructure app only shows a fault if there is a problem in the actual docker infrastructure, rather than if an app using that infrastructure failed to deploy properly.


grkvlt:
@aledsage Thanks, this LGTM with the Brooklyn fix, just testing a bit more locally. The Brooklyn changes also look Ok to merge, I think

mesos/src/main/java/brooklyn/entity/mesos/framework/marathon/MarathonFrameworkImpl.java
```diff
@@ -185,61 +191,42 @@ public void start(Collection<? extends Location> locs) {
...
-
-    @Override
-    public void rebind() {
-        super.rebind();
-
-        // Reload our location definition on rebind
-        ManagementContext.PropertiesReloadListener listener = sensors().get(Attributes.PROPERTIES_RELOAD_LISTENER);
-        if (listener != null) {
-            listener.reloaded();
+        String locationName = config().get(LOCATION_NAME);
```
grkvlt: This will always get the parent configuration, so creating the location will fail. I have a fix that replicates the naming logic that was there originally, which I will commit after merging these changes.

grkvlt:
I am going to merge this to allow building with the merged Brooklyn https://github.com/apache/brooklyn-server/pull/59 and commit a separate fix for Marathon location names.

Issue #264, opened 2016-03-27
NPE when deploying to Apache Mesos cluster location without public address (private VPN)


johnmccabe:
Deploying an app to a mesos cluster fails when that cluster has private addresses only (it works if I add floating IPs - I'm using IBMs Blue Box Openstack here).

The `targetMachine.getHostname()` call in `MarathonPortForwarder.openPortForwarding` is returning null causing the failure.

I've a hacky fix that I'll raise in a PR, have checked it still works with a cluster deployed with public addresses not sure if theres more to it at this point tho.

```
#L17-24Failed after 51.3s: Error invoking start at BasicApplicationImpl{id=HvbuPYmy}: Error invoking start at TomcatServerImpl{id=AatCAQ93}: NoMachinesAvailableException: Failed to start Marathon task

org.apache.brooklyn.core.mgmt.internal.EffectorUtils$EffectorCallPropagatedRuntimeException: Error invoking start at BasicApplicationImpl{id=HvbuPYmy}: Error invoking start at TomcatServerImpl{id=AatCAQ93}: NoMachinesAvailableException: Failed to start Marathon task
	at org.apache.brooklyn.core.mgmt.internal.EffectorUtils$EffectorCallPropagatedRuntimeException.propagate(EffectorUtils.java:299)
	at org.apache.brooklyn.core.mgmt.internal.EffectorUtils$EffectorCallPropagatedRuntimeException.access$100(EffectorUtils.java:266)
	at org.apache.brooklyn.core.mgmt.internal.EffectorUtils.handleEffectorException(EffectorUtils.java:306)
	at org.apache.brooklyn.core.mgmt.internal.EffectorUtils.invokeMethodEffector(EffectorUtils.java:260)
	at org.apache.brooklyn.core.effector.MethodEffector.call(MethodEffector.java:149)
	at org.apache.brooklyn.core.entity.trait.Startable$StartEffectorBody.call(Startable.java:56)
	at org.apache.brooklyn.core.entity.trait.Startable$StartEffectorBody.call(Startable.java:50)
	at org.apache.brooklyn.core.effector.EffectorTasks$EffectorBodyTaskFactory$1.call(EffectorTasks.java:82)
	at org.apache.brooklyn.util.core.task.DynamicSequentialTask$DstJob.call(DynamicSequentialTask.java:359)
	at org.apache.brooklyn.util.core.task.BasicExecutionManager$SubmissionCallable.call(BasicExecutionManager.java:518)
[...]

PR #265, opened 2016-03-27
fix npe when no public address for MarathonTaskLocation.targetMachine

johnmccabe:
Fixes the issue raised in #264, have tested in Blue Box VPN in both a publically accessible cluster and one accessible only via VPN.

@grkvlt I don't have a solid grasp on the handling of hostnames in this case so suspect this may be too simplistic a fix.

### DIFF
--- a/mesos/src/main/java/brooklyn/entity/mesos/framework/marathon/MarathonPortForwarder.java
+++ b/mesos/src/main/java/brooklyn/entity/mesos/framework/marathon/MarathonPortForwarder.java
@@ -18,6 +18,7 @@ package brooklyn.entity.mesos.framework.marathon;
 import java.util.List;
 import java.util.Map;
 
+import brooklyn.location.mesos.framework.marathon.MarathonTaskLocation;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


mesos/src/main/java/brooklyn/entity/mesos/framework/marathon/MarathonPortForwarder.java
```diff
@@ -149,7 +150,7 @@ public HostAndPort openPortForwarding(HasNetworkAddresses targetMachine, int tar
             publicSide = HostAndPort.fromParts(marathonHostname, targetPort);
         }
         pfw.associate(marathonHostname, publicSide, (Location) targetMachine, targetPort);
-        portmap.put(publicSide, HostAndPort.fromParts(targetMachine.getHostname(), targetPort));
+        portmap.put(publicSide, HostAndPort.fromParts(((MarathonTaskLocation) targetMachine).getSubnetHostname(), targetPort));
```
grkvlt: Maybe do something like:

```
MarathonTaskLocation location = (MarathonTaskLocation) targetMachine;
pfw.associate(marathonHostname, publicSide, targetMachine, targetPort);
String hostname = Optional.fromNullable(location.getHostname()).or(location.getSubnetHostname());
portmap.put(publicSide, HostAndPort.fromParts(hostname, targetPort));
```

grkvlt:
@johnmccabe  @aledsage Are you sure this is right?

PR #269, opened 2016-03-30
Fix mesos rebind and no entities for unmanaged tasks

aledsage:


mesos/src/main/java/brooklyn/entity/mesos/MesosSlaveImpl.java
```diff
@@ -70,6 +70,7 @@ public void init() {
 
         EnricherSpec<?> serviceUp = Enrichers.builder()
                 .propagating(ImmutableMap.of(SLAVE_ACTIVE, SERVICE_UP))
+                .suppressDuplicates(true)
```
grkvlt: @aledsage I didn't know about this - very useful

grkvlt:
@aledsage LGTM

PR #272, opened 2016-04-01
GroupPlacementStrategy update and various other fixes

grkvlt:
- Fix issues with `GroupPlacementStrategy` using `Monitor` mutexes
- Add extra command support to `VanillaDockerApplication`
- Better `docker exec` support for `useSsh: false`
- Use `docker cp` instead of `scp` for  `useSsh: false`
- Support `volumesFrom` container config
- Copies of files for https://github.com/jclouds/jclouds-labs/pull/253


grkvlt:
@aledsage @nakomis can you take a look at this please?

grkvlt:
Note we will probably still need to make a private jclouds 1.9.3 build for Clocker to use the new template option...

docker/src/main/java/brooklyn/location/docker/DockerContainerLocation.java
```diff
@@ -268,13 +277,75 @@ private void parseDockerCallback(String commandString) {
     }
 
     @Override
+    public int copyTo(final Map<String,?> props, final InputStream src, final long filesize, final String destination) {
+        return copyTo(props, src, destination);
+    }
+
+    @Override
+    public int copyTo(final Map<String,?> props, final InputStream src, final String destination) {
+        File tmp = null;
```
nakomis: Why not use `super.copyTo()` if SSH available?

docker/src/main/java/brooklyn/location/docker/strategy/GroupPlacementStrategy.java
```diff
@@ -70,4 +137,68 @@ public boolean apply(DockerHostLocation input) {
             }
         }
     }
+
+    /**
+     * Look up the {@link monitor} for an application.
```
nakomis: Should be `Monitor`, not `monitor`

docker/src/main/java/brooklyn/location/docker/strategy/GroupPlacementStrategy.java
```diff
@@ -70,4 +137,68 @@ public boolean apply(DockerHostLocation input) {
...
+     */
+    protected Monitor lookupMonitor(String mutexId) {
+        synchronized (lock) {
+            Map<String, Monitor> map = getDockerInfrastructure().config().get(MONITOR_MAP);
+            return (map != null) ? map.get(mutexId) : null;
+        }
+    }
+
+    /**
+     * Create a new {@link monitor} and optionally the {@link #monitor_MAP map}
```
nakomis: `Monitor`, `MONITOR_MAP`

patches/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
```diff
@@ -0,0 +1,650 @@
...
+              equal(this.cpuShares, that.cpuShares) &&
+              equal(this.env, that.env) &&
+              equal(this.portBindings, that.portBindings) &&
+              equal(this.extraHosts, that.extraHosts) &&
+              equal(this.volumesFrom, that.volumesFrom);
+   }
+
+   @Override
+   public int hashCode() {
+      return Objects.hashCode(super.hashCode(), volumes, hostname, dns, memory, entrypoint, commands, cpuShares, env, portBindings, extraHosts);
```
nakomis: Missing `volumesFrom`

patches/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
```diff
@@ -0,0 +1,650 @@
...
+
+   public List<String> getEnv() { return env; }
+
+   public Map<Integer, Integer> getPortBindings() { return portBindings; }
+
+   public String getNetworkMode() { return networkMode; }
+
+   public Map<String, String> getExtraHosts() { return extraHosts; }
+
+   public static class Builder {
```
nakomis: Not sure this is necessary as `DockerTemplateOptions` already has fluent methods, and the `Builder` methods below simple return a DTO, not a builder

nakomis:
Would be useful if the `patches` classes contained a description of what has been changed, and a link to any relevant PR

nakomis:
A couple of minor comments, but other than that, LGTM

patches/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
grkvlt: Is last item. These have been reviewed when merged into jclouds-labs, and are just copies...

patches/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
grkvlt: Yes, there are plans to update the builder to use auto-values in another jclouds-labs PR...

grkvlt:
Good call on documenting patches project content...

grkvlt:
Will fix typos and merge...

patches/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
grkvlt: Oh, right. It's not in `hashCode` - can't fix that here, will maybe get it in with the AutoValue builder PR

grkvlt:
Waiting on review comments from @aledsage before merge

johnmccabe:
Entrypoint looking better, I was able to successfully deploy the following:
```yaml
location: dc
services:
- type: docker:mariadb:5.5
  env:
    MYSQL_ROOT_PASSWORD: "helooooooo"
```
But the following failed, with the on-fire container not getting removed when the app is stopped:
```yaml
location: dc
services:
- type: docker:mariadb:5.5
  id: db
  docker.container.name: mysql
  docker.image.entrypoint:
  - /docker-entrypoint.sh
  env:
    MYSQL_ROOT_PASSWORD: "helooooooo"
```
```
Required entity not healthy: VanillaDockerApplicationImpl{id=H1QpmZBt}
Failure running task start (iXjyKWq5): Error invoking start at VanillaDockerApplicationImpl{id=H1QpmZBt}: Error invoking start at DockerContainerImpl{id=ecFntZQN}: RunNodesException: error running 1 node group(brooklyn-o5n62j-johnmccabe) location(docker) image(4898c21213910b7c92f584e44f5f23a02317fa4fd0e0eb3977d5a89a9210e52a) size(small) options(DockerTemplateOptions{dns=[], hostname=mysql, memory=null, cpuShares=null, entrypoint=[/docker-entrypoint.sh], commands=null, volumes={}, env=[MYSQL_ROOT_PASSWORD=marmalad3], portBindings={}, extraHosts={}, volumesFrom=[]}) Execution failures: 0 error[s] Node failures: 1) IllegalStateException on node 841766ace18a608f35f5dec04a33764a638152bb51ddd5def0ae7dfa0eb0d001: java.lang.IllegalStateException: node(841766ace18a608f35f5dec04a33764a638152bb51ddd5def0ae7dfa0eb0d001) terminated at org.jclouds.compute.functions.PollNodeRunning.apply(PollNodeRunning.java:76) at org.jclouds.compute.functions.PollNodeRunning.apply(PollNodeRunning.java:42) at org.jclouds.compute.strategy.CustomizeNodeAndAddToGoodMapOrPutExceptionIntoBadMap.call(CustomizeNodeAndAddToGoodMapOrPutExceptionIntoBadMap.java:117) at org.jclouds.compute.strategy.CustomizeNodeAndAddToGoodMapOrPutExceptionIntoBadMap.apply(CustomizeNodeAndAddToGoodMapOrPutExceptionIntoBadMap.java:142) at org.jclouds.compute.strategy.CustomizeNodeAndAddToGoodMapOrPutExceptionIntoBadMap.apply(CustomizeNodeAndAddToGoodMapOrPutExceptionIntoBadMap.java:49) at com.google.common.util.concurrent.Futures$1.apply(Futures.java:713) at com.google.common.util.concurrent.Futures$ChainingListenableFuture.run(Futures.java:861) at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) at java.lang.Thread.run(Thread.java:745) 1 error[s]
```
Also noted that when deploying 2 linked apps the second isn't being provisioned.

I suggest we deal with those in a later PR rather than drag this one out with unrelated issues.

grkvlt:
@johnmccabe OK, integration tests passing, any objection to merging now?

dist/src/main/assembly/conf/brooklyn/default.catalog.bom
```diff
@@ -30,11 +30,11 @@ brooklyn.catalog:
         docker.policy.ha.enable: false
         docker.container.strategies:
         - $brooklyn:object:
-            type: clocker.docker.location.strategy.MaxContainersPlacementStrategy
+            type: clocker.docker.location.strategy.basic.MaxContainersPlacementStrategy
```
aledsage: This kind of change breaks backwards compatibility of persisted state - we need to be increasingly careful of that. One solution is to use the Brooklyn `DeserializingClassRenamesProvider`. We can add to the `classpath://org/apache/brooklyn/core/mgmt/persist/deserializingClassRenames.properties`.

Perhaps we want to improve Brooklyn so that it's not just a single file, but instead we also pick up all properties files in some well-defined classpath folder.

dist/src/main/assembly/conf/logback.xml
```diff
@@ -16,7 +16,9 @@
 -->
 <configuration>
     <property name="logging.basename" scope="context" value="brooklyn-clocker" />
-    <logger name="brooklyn.clocker" level="DEBUG" />
+
+    <logger name="clocker" level="DEBUG" />
```
aledsage: Good to update AMP with this as well :-)

docker/src/main/java/clocker/docker/location/DockerContainerLocation.java
```diff
@@ -269,13 +273,100 @@ private void parseDockerCallback(String commandString) {
...
+        Map<String,?> nonPortProps = Maps.filterKeys(props, Predicates.not(Predicates.containsPattern("port")));
+        boolean entitySsh = Boolean.TRUE.equals(entity.config().get(DockerContainer.DOCKER_USE_SSH));
+        boolean dockerSsh = Boolean.TRUE.equals(getOwner().config().get(DockerContainer.DOCKER_USE_SSH));
+        if (entitySsh && dockerSsh) {
+            return super.copyTo(nonPortProps, src, destination);
+        } else {
+            File tmp = null;
+            try {
+                tmp = File.createTempFile(dockerContainer.getId(), Urls.getBasename(destination));
+                hostMachine.copyTo(nonPortProps, src, tmp.getAbsolutePath());
```
aledsage: This seems wrong: we create a temp file on the *Clocker* server, and then we use that file's path as the target on the *docker host*. I can see that if the Clocker and docker host are similarly configured, then this will work. But if for example someone tried to run Clocker on windows (or on OS X with a different default tmp directory) then the path would not be applicable on the docker host. Why don't we just construct a name like `org.apache.brooklyn.util.os.Os.mergePathsUnix("/tmp", dockerContainer.getId() + Urls.getBasename(destination))` (possibly with some random id as well)?

Same for the other `copyTo` method below.

docker/src/main/java/clocker/docker/location/DockerContainerLocation.java
```diff
@@ -188,12 +190,13 @@ public int execScript(Map<String,?> props, String summaryForLogging, List<String
         for (String commandString : filtered) {
             parseDockerCallback(commandString);
         }
-        if (getOwner().config().get(DockerContainer.DOCKER_USE_SSH)) {
+        boolean entitySsh = Boolean.TRUE.equals(entity.config().get(DockerContainer.DOCKER_USE_SSH));
+        boolean dockerSsh = Boolean.TRUE.equals(getOwner().config().get(DockerContainer.DOCKER_USE_SSH));
+        if (entitySsh && dockerSsh) {
```
aledsage: Why do we need it set on both the entity (i.e. the app component being deployed) and the owner (i.e. on the container)? What will set it on the container? Do you expect that to be explicitly set in the Clocker blueprint, or something more dynamic?

docker/src/main/java/clocker/docker/entity/SoftLayerHostnameCustomizer.java
```diff
@@ -0,0 +1,63 @@
...
+        return new SoftLayerHostnameCustomizer();
+    }
+
+    @Override
+    public void customize(JcloudsLocation location, ComputeService computeService, JcloudsMachineLocation machine) {
+        JcloudsSshMachineLocation ssh = (JcloudsSshMachineLocation) machine;
+        String name = ssh.getOptionalNode().get().getName();
+
+        List<String> commands = ImmutableList.of(
+                BashCommands.sudo("rm -f /etc/hostname"),
```
aledsage: This looks extreme! I don't get how it works (or what it's supposed to do). It doesn't use `name` for anything except the subsequent log statement that says it set the hostname to that value.

Could you add more javadoc to the class to say what the SoftLayer hostname problem is, and what the fix is?

docker/src/main/java/clocker/docker/entity/container/DockerContainerImpl.java
```diff
@@ -420,13 +425,20 @@ private DockerTemplateOptions getDockerTemplateOptions() {
         }
         options.env(env);
 
-        // Entrypoint
+        // Entrypoint and commands
+        // TODO parse string into list?
```
aledsage: See Brooklyn's `TypeCoercions`, which does `JavaStringEscapes.unwrapJsonishListIfPossible(input)`. Would that be useful? But probably only worth adding something like that if really is needed.

docker/src/main/java/clocker/docker/entity/container/application/VanillaDockerApplicationSshDriver.java
```diff
@@ -63,6 +65,16 @@ public void install() {
     }
 
     @Override
+    public void copyRuntimeResources() {
+        Map<String, String> runtimeFiles = entity.getConfig(VanillaSoftwareProcess.RUNTIME_FILES);
+        Map<String, String> runtimeTemplates = entity.getConfig(VanillaSoftwareProcess.RUNTIME_TEMPLATES);
+        if ((runtimeFiles != null && runtimeFiles.size() > 0) ||
+                (runtimeTemplates != null && runtimeTemplates.size() > 0)) {
+            super.copyRuntimeResources();
```
aledsage: Should we change Brooklyn's `AbstractSoftwareProcessDriver`, so that it only does the `createDirectory(getRunDir(), "create run directory")` if there are runtime resources to copy? (I presume nothing downstream in normal Brooklyn entities relies on runDir existing because of `copyRuntimeResources`, but we'd need to check that).

dist/src/main/assembly/conf/brooklyn/default.catalog.bom
grkvlt: Allowing each jar or project to provide a file with a well-known or discoverable name for a `DeserializingClassRenamesProvider` would be good. I can implement one for the (large amount of) Clocker renaming that was done for the OSGi work, but in another PR.

dist/src/main/assembly/conf/logback.xml
grkvlt: Yes, was wondering why I didn't get any logging after the package renaming, and realised I needed to add a new logger...

docker/src/main/java/clocker/docker/location/DockerContainerLocation.java
grkvlt: No, the `File` object is just a representation. We are then copying the source to that temp file on the Docker host like you suggest. The usage is a bit misleading, I agree. I got the implementation wrong myself the first time, because of that 😸

docker/src/main/java/clocker/docker/location/DockerContainerLocation.java
grkvlt: Yes, it might be set on the Clocker blueprint itself.

docker/src/main/java/clocker/docker/entity/SoftLayerHostnameCustomizer.java
grkvlt: SoftLayer seemed to be setting the hostname to `(none)` which was annoying. I can document further, though.

docker/src/main/java/clocker/docker/location/strategy/util/StrategyPredicates.java
```diff
@@ -0,0 +1,133 @@
...
+    /** @see {@link ChildrenOfPredicate} */
+    public static Predicate<DockerHostLocation> childrenOf(Entity parent) {
+        return new ChildrenOfPredicate(parent);
+    }
+
+    /**
+     * A {@link Predicate} that checks if a {@link DockerHostLocation host} contains
+     * entities that are children of the given parent.
+     */
+    public static class ChildrenOfPredicate implements Predicate<DockerHostLocation> {
```
aledsage: Personal preference to not have the classes being public (e.g. in Brooklyn's `EntityPredicates` then the classes are protected). We have the static methods like `childrenOf` to create them instead.

But if you think it's likely that others will want to extend `ChildrenOfPredicates` (or at least if you think it's worth supporting, versus keeping our API/SPI smaller) then that's fine.

docker/src/main/java/clocker/docker/entity/container/DockerContainerImpl.java
grkvlt: Yes, probably good to support coercion of `String` to a `List`, but would need to understand shell quoting, since splitting on spaces is not going to work with e.g. `echo "Hello, World!"` (returns three tokens)

docker/src/main/java/clocker/docker/entity/container/application/VanillaDockerApplicationSshDriver.java
grkvlt: Yes, we should, but I didn't want to have to wait to get the change into Brooklyn.

docker/src/main/java/clocker/docker/location/strategy/util/StrategyPredicates.java
grkvlt: Reason for having them public is not for extension, but so that they can be referenced by class name in a config key if that is needed, e.g. you have a key that is a `List<Predicate<Entity>>` for filtering. Otherwise they can only be used in code via the helper methods.

grkvlt: In fact, I'd actually suggest the inner classes in `EntityPredicates` should be made public too, for the same reason.

docker/src/main/java/clocker/docker/location/strategy/basic/GroupPlacementStrategy.java
```diff
@@ -0,0 +1,182 @@
...
+
+        // Check if hosts have any deployed entities that share a parent with the input entity
+        Optional<DockerHostLocation> sameParent = Iterables.tryFind(sameApplication, childrenOf(entity.getParent()));
+        if (sameParent.isPresent()) {
+            LOG.debug("Returning {} (same parent) for {} placement", sameParent.get(), entity);
+            return ImmutableList.copyOf(sameParent.asSet());
+        }
+
+        // Remove hosts if they do not have any entities from our application deployed there
+        Iterables.removeIf(available, hasApplicationId(entity.getApplicationId()));
```
aledsage: This line of code seems to do the opposite to the comment: will it not remove hosts if they *do* have any entities from our application deployed there? I presume that's what you are intending, so that things with a different parent in the same app are mutually exclusive.

docker/src/main/java/clocker/docker/location/strategy/basic/GroupPlacementStrategy.java
```diff
@@ -0,0 +1,182 @@
...
+import org.apache.brooklyn.util.collections.MutableList;
+import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.core.flags.SetFromFlag;
+import org.apache.brooklyn.util.core.mutex.WithMutexes;
+import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.apache.brooklyn.util.time.Duration;
+
+/**
+ * A {@link DockerAwarePlacementStrategy strategy} that requires entities with
+ * the same parent to use the same host.
```
aledsage: It also requires that entities with a different parent (in the same app) are on a different host.

docker/src/main/java/clocker/docker/location/strategy/basic/GroupPlacementStrategy.java
```diff
@@ -0,0 +1,182 @@
...
+    public List<DockerHostLocation> filterLocations(List<DockerHostLocation> locations, Entity entity) {
+        if (locations == null || locations.isEmpty()) {
+            return ImmutableList.of();
+        }
+        if (getDockerInfrastructure() == null) config().set(DOCKER_INFRASTRUCTURE, Iterables.getLast(locations).getDockerInfrastructure());
+        List<DockerHostLocation> available = MutableList.copyOf(locations);
+        boolean requireExclusive = config().get(REQUIRE_EXCLUSIVE);
+
+        try {
+            acquireMutex(entity.getApplicationId(), "Filtering locations for " + entity);
```
aledsage: I find it confusing the way that `releaseMutex()` is called in the `DockerLocation` code. Deserves a comment here to explain that. Not until you search for `releaseMutex` does one understand how this code can work.

docker/src/main/java/clocker/docker/location/DockerLocation.java
```diff
@@ -245,13 +239,20 @@ public MachineLocation obtain(Map<?,?> flags) throws NoMachinesAvailableExceptio
...
             LOG.debug("Obtain a new container from {} for {}", machine, entity);
+            Map<?,?> hostFlags = MutableMap.copyOf(flags);
+            DockerContainerLocation container = machine.obtain(hostFlags);
+            containers.put(machine, container.getId());
+            return container;
+        } finally {
+            // Release any placement strategy locks
+            for (DockerAwarePlacementStrategy strategy : Iterables.concat(strategies, entityStrategies)) {
+                if (strategy instanceof WithMutexes) {
+                    ((WithMutexes) strategy).releaseMutex(entity.getApplicationId());
```
aledsage: This try-finally block in `DockerLocation` is surprising. After calling `strategy.filterLocations`, there is a big chunk of code not guarded by the finally block, so if there is an exception during that work, the acquired mutex will not be released.

aledsage: My gut feel (not fully thought through) was that if `strategy.filterLocations` fails to find a machine then we'd provision a new machine and then ask the strategy again which machine it wanted. The plus side of that would be that we don't need to keep holding the lock for a long time. The downsides are: 1) if we're not holding the lock, then then new machine could get grabbed by another app so our original obtain request might get starved; and 2) if the strategy keeps saying "no" then we'd keep getting more machines (but we'd guard against that by failing after a certain number of attempts, perhaps?).

I'm fine with leaving it as-is. But worth some more discussion.

docker/src/main/java/clocker/docker/location/strategy/basic/GroupPlacementStrategy.java
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * Can be configured to require exclusive use of the host with the
+ * {@link #REQUIRE_EXCLUSIVE exclusive} ({@code docker.constraint.exclusive})
+ * option set to {@code true}; normally {@code false}.
+ *
+ * @since 1.1.0
+ */
+@Beta
+@ThreadSafe
+public class GroupPlacementStrategy extends AbstractDockerPlacementStrategy implements WithMutexes {
```
aledsage: I wonder if `GroupPlacementStrategy` is the best name. Perhaps `SiblingPlacementStrategy`? "Group" is a very generic term, and in Brooklyn has a more specific meaning that is not applicable here.

aledsage:
Thanks @grkvlt - looks good. I've gone through the code (some in detail, other bits just skimming). I've included some comments but nothing major. Over to you to see what you think of the comments and then to merge.

docker/src/main/java/clocker/docker/entity/container/application/VanillaDockerApplicationSshDriver.java
aledsage: +1; can you raise a PR in Brooklyn for that at some point soon'ish please.

docker/src/main/java/clocker/docker/location/DockerContainerLocation.java
aledsage: You say it might be set on the clocker blueprint itself: but it seems that you *require* it to be set on the clocker blueprint itself, otherwise `entitySsh && dockerSsh` will not both be true. Is that right?

docker/src/main/java/clocker/docker/location/DockerContainerLocation.java
aledsage: That's how I read the code (except we have also create a temp file on the clocker server because that is what `File.createTempFile` does). I have two concerns:
1. the path to the tmp file that was sensible on the clocker server may be entirely unsuitable on the docker host, causing it to fail;
2. the file on the docker host is not a temporary file - nothing deletes it. That could be a security issue, depending on the nature of the file being copied.

dist/src/main/assembly/conf/brooklyn/default.catalog.bom
aledsage: I think we can ignore Clocker renames of a certain age (not sure exactly what, but OSGi work is probably old enough). It's the more recent things that concern me (particularly where potential customers etc have been pointed at the download url, as well as the more forgiving community at large).

grkvlt: No, the OSGi package rename was only a couple days ago, and involved a **lot** of changes. Things like: `brooklyn.clocker.docker.entity.container` getting split up and becoming `clocker.docker.entity`,  `clocker.docker.entity.container`,  `clocker.docker.entity.container.application`, `clocker.docker.entity.container.registry` and  `clocker.docker.entity.microservice`. Same for the Mesos packages, and so on. The placement strategy change here was just building on that.

docker/src/main/java/clocker/docker/location/DockerContainerLocation.java
grkvlt: OK, good point. I'll make some changes so that the right thing happens, and add comments to explain what is actually happening.

docker/src/main/java/clocker/docker/location/DockerContainerLocation.java
grkvlt: The default value of that config key is `Boolean.TRUE` so we are _really_ checking if either of the application or the Clocker blueprints have set it to `false`... Maybe a comment to that effect would be useful?

docker/src/main/java/clocker/docker/location/strategy/basic/GroupPlacementStrategy.java
grkvlt: Yes, will fix comment.

docker/src/main/java/clocker/docker/location/strategy/basic/GroupPlacementStrategy.java
grkvlt: Probably needs documented both here and in the `obtain()` method.

docker/src/main/java/clocker/docker/location/DockerLocation.java
grkvlt: Yes, probably needs more testing and thought given to non-happy-path cases. The lock entry has a timeout to prevent blocking forever, but still not optimal.

docker/src/main/java/clocker/docker/location/strategy/basic/GroupPlacementStrategy.java
grkvlt: Maybe, it was meant in the sense of grouping entities together, really. Not sure about sibling either though. Probably doesn't matter as long as we document carefully, both here and on the [wiki](https://github.com/brooklyncentral/clocker/wiki/PlacementStrategy)

docker/src/main/java/clocker/docker/location/DockerContainerLocation.java
aledsage: Ah, I missed that default value when just looking at the PR changes - sounds good.

grkvlt:
Tested on various clouds successfully by @aledsage @johnmccabe and @grkvlt

PR #279, opened 2016-04-15
pull security group from location (requires #272)

johnmccabe:
This requires #272 to be merged.

If the `docker.host.securityGroup` is omitted then attempt to pull the securityGroup string from the location when its defined unambiguously in the location, ie a String.

Currently does not attempt to pull the securityGroup from the location when an ArrayList or Iterator, multiple securityGroup support isn't currently implemented in Clocker (should it?).


johnmccabe:
fyi @grkvlt

grkvlt:
Included in #272 now, thanks @johnmccabe

PR #285, opened 2016-04-20
Fix behaviour when stopping containers

grkvlt:
Ensures `DockerContainerImpl#stop()` is called and that `docker rm` gets run on the container.


johnmccabe:
apologies @grkvlt only just got round to testing this, it looks good, thanks!

Issue #294, opened 2016-05-03
1.1.0 clocker release breaks AWS docker weave cloud.


avedz:
When a new docker image is creating in a weave docker cloud this error appears:
No such plugin error

grkvlt:
@avedz thanks for the report. I'll check if we have the right plugin details for libnetwork and weave. I thought I had changed them. Look out for a 1.1.1 release with this fixed, shortly.

grkvlt:
@avedz This is now fixed in master by #296

PR #296, opened 2016-05-04
Weave plugin fix

grkvlt:
Fixes issue in #294 caused by incorrect weave startup commands still using old proxy.


grkvlt:
Tested with Weave 1.5.0

PR #329, opened 2016-09-30
Kubernetes version update

grkvlt:
Fixes issue with non-reporting of external IP addresses, updates to stable versions of components, adds Prometheus monitoring solution and corrects BOM file catalog contents so they work with classic Brooklyn.


m4rkmckenna:
@grkvlt If we are changing the artifact ids of the modules [common, swarm, kubernetes] please rename the folders also

common/pom.xml
```diff
@@ -2,65 +2,74 @@
...
 
-  <parent>
-    <groupId>io.brooklyn.clocker</groupId>
-    <artifactId>clocker-parent</artifactId>
-    <version>2.0.0-SNAPSHOT</version> <!-- CLOCKER_VERSION -->
-  </parent>
+    <parent>
+        <groupId>io.brooklyn.clocker</groupId>
+        <artifactId>clocker-parent</artifactId>
+        <version>2.0.0-SNAPSHOT</version> <!-- CLOCKER_VERSION -->
```
aledsage: We are now on 2.1.0-SNAPSHOT in master.

common/pom.xml
```diff
@@ -2,65 +2,74 @@
...
+    </parent>
 
-  <name>Clocker :: Common</name>
-  <artifactId>common</artifactId>
-  <packaging>jar</packaging>
+    <name>Clocker :: Common</name>
+    <artifactId>clocker-common</artifactId>
```
aledsage: Why rename these to "clocker-common", rather than just "common"?

common/src/main/resources/docker/docker.bom
```diff
@@ -357,6 +364,46 @@ brooklyn.catalog:
             command: |
               cat ${DOCKER_CERT_PATH}/csr.pem
 
+  - id: docker-engine-with-resilience
```
aledsage: Whoever is reviewing more thoroughly: can you diff this text from that removed from `swarm/src/main/resources/swarm/swarm.bom` and let us know if anything has changed (ignoring white-space). And paste the diff as a response to this comment.

pom.xml
```diff
@@ -46,6 +46,17 @@
         </license>
     </licenses>
 
+    <developers>
```
aledsage: Conflicts with `developers` section added separately, to have sufficient metadata for deploying 2.0.0 to sonatype. Very happy for you to update the existing one (including role "founder" obviously).

kubernetes/src/main/resources/kubernetes/catalog.bom
```diff
@@ -0,0 +1,5 @@
+brooklyn.catalog:
+  items:
+    - classpath://docker/catalog.bom
```
aledsage: How does this work outside of Brooklyn classic? This is in a different module, so a different OSGi bundle. Would we not need the bundleId in the URL as well?

grkvlt:
Replaced by #346 and #348

PR #336, opened 2016-10-03
Update README.md

andreaturli:


drigodwin:
LGTM and the URLs are resolving now so merging.

PR #341, opened 2016-10-03
Improve README

aledsage:


andreaturli:
thx @aledsage I'd probably keep also the instruction for developers to test clocker using the latest artifacts, i.e.:
https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=io.brooklyn.clocker&a= common&v=LATEST&e=jar

PR #353, opened 2016-10-12
rerun sensors script

duncangrant:
DO NOT MERGE


duncangrant:
Mark has found a better solution

PR #355, opened 2016-10-13
install curl with docker engine

duncangrant:
A number of the blueprints that depend on docker-engine
require curl.  So far it has been assumed to be available
but this is probably not safe.


duncangrant:
@grkvlt I added this as I wanted to use curl in a separate blueprint and I couldn't find an available pre or post install command.  I don't really like adding it where I have but I hit the same issue if I tried to add to centos-software-process as well.

duncangrant:
I'm going to close this until I come up with a better place to put this

PR #356, opened 2016-10-13
Fixes for various issues spotted during testing

Graeme-Miller:


common/src/main/resources/docker/docker.bom
```diff
@@ -433,6 +433,10 @@ brooklyn.catalog:
 
         launch.command: |
           docker run -d ${IMAGE_DETAILS}
+          
+        checkRunning.command: |  
+          result="$(docker ps | grep ${IMAGE_DETAILS} | wc -l)"
+          if [ $result -ne 0 ]; then exit 0; else exit 1; fi;
```
neykov: See how `docker-vm-container` implements this, could re-use.

common/src/main/resources/docker/docker.bom
```diff
@@ -509,8 +513,8 @@ brooklyn.catalog:
           }
           rm -f "${PID_FILE}" # docker won't overwrite
           DOCKER="docker run -d --cidfile \"${PID_FILE}\" --net=host"
-          [ "${DOCKER_RUN_VOLUMES}" ] && DOCKER="${DOCKER} $( parse_docker_volumes )"
-          [ "${DOCKER_RUN_ENV}" ] && DOCKER="${DOCKER} $( parse_docker_env )"
+          [[ $DOCKER_RUN_VOLUMES != [] ]] && DOCKER="${DOCKER} $( parse_docker_volumes )"
+          [[ $DOCKER_RUN_ENV != {} ]] && DOCKER="${DOCKER} $( parse_docker_env )"
```
neykov: Could any of those still be the empty string in which case the check will be a success?

Graeme-Miller: I don't believe so, these are set from a DSL brooklyn:config call.
There are tests which cover the config being empty and the config being full- they both pass with this change.

common/src/main/resources/docker/docker.bom
Graeme-Miller: good point, but I wanted to check that the specific image was running.
Do you think it is ok to merge without?

neykov:
Can be merged after resolving merge conflicts.

PR #357, opened 2016-10-13
Swarm manager load balancer fix

Graeme-Miller:
Added a latch, to prevent the swarm managers from starting until the load balancer has started.
This is because the load balancer is only re-configured when a swarm manager is added to the cluster- so we need to make sure the balancer is up first.

Also added an explicit CD to the RUN_DIR when we are configuring swarm

### DIFF
--- a/swarm/src/main/resources/swarm/swarm.bom
+++ b/swarm/src/main/resources/swarm/swarm.bom
@@ -253,7 +253,8 @@ brooklyn.catalog:
                 shell.env:
                   SWARM_ENDPOINTS:
                     $brooklyn:attributeWhenReady("swarm.endpoints")
-                command: |
+                command: |                  
+                  cd ${RUN_DIR}
                   cat > servers.conf <<-EOF
                   backend servers
                     balance roundrobin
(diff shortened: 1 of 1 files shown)


swarm/src/main/resources/swarm/swarm.bom
```diff
@@ -253,7 +253,8 @@ brooklyn.catalog:
                 shell.env:
                   SWARM_ENDPOINTS:
                     $brooklyn:attributeWhenReady("swarm.endpoints")
-                command: |
+                command: |                  
+                  cd ${RUN_DIR}
```
Graeme-Miller: I have observed this command being called when we weren't in the RUN_DIR. This resulted in the servers.conf file being created in home. 

I am not 100% sure this is needed, but I think it is safer to keep it in.

duncangrant: Could you use executionDir?  I can't see how - I'm just speculating?

duncangrant: Actually presumably executionDir could be ${RUN_DIR}?

PR #359, opened 2016-10-14
Adds an effector to swarm-manager that interacts with the docker CLI

m4rkmckenna:


drigodwin (changes requested):
Looks good but probably needs some tests 👍

swarm/src/main/resources/swarm/swarm.bom
```diff
@@ -601,6 +601,17 @@ brooklyn.catalog:
               DOCKER_HOST: $brooklyn:attributeWhenReady("swarm.url")
             command: |
               docker info | grep "^Containers:" | sed -e "s/^Containers: //"
+        - type: org.apache.brooklyn.core.effector.ssh.SshCommandEffector
```
aledsage: Looking at the other `SshCommandSensor`s which run `docker info`, they include the following:

```
            shell.env:
              DOCKER_HOST: $brooklyn:attributeWhenReady("swarm.url")
```

I think you probably need that here as well, for the `docker` ssh command to work.

aledsage:
@m4rkmckenna I agree with @drigodwin - would be good to have a test with this before merging, so we can test it automatically in the QA Framework.

PR #361, opened 2016-10-17
Add ECS ECS clustering for Docker

grkvlt:
Creates a cluster of Docker Engines for use with Amazon ECS, running the ECS agent to register with the container service. Scaling and service replacement is managed by Brooklyn.


grkvlt:
The screenshot below shows what an ECS cluster managed by Brooklyn looks like in the AWS console, with one task running.

ecs-console

common/src/main/resources/docker/ecs.bom
```diff
@@ -0,0 +1,231 @@
...
+            The name of the ECS cluster
+          type: string
+          default: "clocker"
+
+      brooklyn.children:
+        - type: docker-cluster
+          id: docker-cluster
+          name: "docker-cluster"
+
+  - id:  docker-cluster
```
aledsage: `docker-cluster` seems a strange name for this, to add to the catalog. We'll have lost the context that it is for EC2 container service. We should probably have "ecs" in the catalog item id.

common/src/main/resources/docker/ecs.bom
```diff
@@ -0,0 +1,231 @@
...
+    name: "EC2 Container Service Docker Cluster"
+    description: |
+      Creates a cluster of Docker engines for use with Amazon ECS
+    itemType: template
+    iconUrl: https://upload.wikimedia.org/wikipedia/commons/thumb/a/af/AWS_Simple_Icons_Compute_ECS.svg/240px-AWS_Simple_Icons_Compute_ECS.svg.png
+    item:
+      services:
+        - type: ecs-cluster
+
+  - id:  ecs-cluster
```
aledsage: It's scary that we're repeating all of these brooklyn.parameters in different entities in this same file.

What is the purpose of having `ecs-cluster` and `docker-cluster` as separate items?

common/src/main/resources/docker/ecs.bom
```diff
@@ -0,0 +1,231 @@
...
+            autoscaler.metricLowerBound: 0.00
+            autoscaler.metricUpperBound:
+              $brooklyn:config("docker.scaling.cpu.limit")
+            autoscaler.minPoolSize:
+              $brooklyn:config("docker.initial.size")
+            autoscaler.maxPoolSize:
+              $brooklyn:config("docker.max.size")
+            autoscaler.resizeUpStabilizationDelay: 30s
+            autoscaler.resizeDownIterationMax: 0 # Disable scaling down
+            autoscaler.resizeDownStabilizationDelay: forever
```
aledsage: Will this cause us to track an infinite history of cluster load, so that we try to calculate what load has been happening "forever"? (I haven't dug into the Brooklyn `AutoScalerPolicy` to check; if you're going to rely on this, then perhaps we should have a test in Brooklyn? Or maybe we should just add an explicit `autoscaler.resizeDownEnabled: false`).

common/src/main/resources/docker/ecs.bom
```diff
@@ -0,0 +1,231 @@
...
+          brooklyn.config:
+            shell.env:
+              ECS_AGENT_VERSION: $brooklyn:config("ecs.agent.version")
+              ECS_CLUSTER_NAME: $brooklyn:config("ecs.cluster.name")
+
+            install.command: |
+              sudo mkdir -p /var/log/ecs
+              sudo mkdir -p /var/lib/ecs/data
+              sudo sysctl -w net.ipv4.conf.all.route_localnet=1
+              sudo iptables -t nat -A PREROUTING -p tcp -d 169.254.170.2 --dport 80 -j DNAT --to-destination 127.0.0.1:51679
```
aledsage: Does this assume CentOS 6? I'd have thought we'd want `firewalld` for CentOS 7.

grkvlt: Based on the ECS instructions for installing the agent

common/src/main/resources/docker/ecs.bom
grkvlt: Interesting, not sure. We should definitely check, as this is part of the Swarm and Kubernetes autoscalers too.

common/src/main/resources/docker/ecs.bom
grkvlt: Agree

grkvlt:
@aledsage what tests would you suggest? The ECS agent will fail on startup if an ECS cluster is not available with the provided name.

common/src/main/resources/docker/ecs.bom
neykov: I see what you mean now. Agree it would be nice to be able to express this case in a more concise way.

grkvlt:
@aledsage I added tests that both check service up and also verify that custom sensors have correct values, this should be good to merge?

PR #362, opened 2016-10-17
Increase etcd default

drigodwin:
The default of `1` can intermittently break, so up it to `3` so the average user will not hit this. A note to fix it has been made elsewhere

### DIFF
--- a/swarm/src/main/resources/swarm/swarm.bom
+++ b/swarm/src/main/resources/swarm/swarm.bom
@@ -53,7 +53,7 @@ brooklyn.catalog:
           description: |
             Size of the etcd cluster when created initially
           type: integer
-          default: 1
+          default: 2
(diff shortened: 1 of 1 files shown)


swarm/src/main/resources/swarm/swarm.bom
```diff
@@ -53,7 +53,7 @@ brooklyn.catalog:
           description: |
             Size of the etcd cluster when created initially
           type: integer
-          default: 1
+          default: 2
```
neykov: Doesn't etcd need an odd number of cluster members? Does it just ignore one of the members is they are even?

drigodwin: I don't think this is the case but it is recommended to use odd numbers - I'll up it to 3.

PR #364, opened 2016-10-19
Formatting and Configuration Updates

grkvlt:
Updated Kubernetes to version 1.4.3, quote labels consistently, and fix configuration for provisioning properties keys `minRam` and `minCores`. Also splits up the `src/main/resources` folders more logically.


kubernetes/pom.xml
```diff
@@ -17,12 +17,17 @@
...
+                <filtering>false</filtering>
+            </resource>
+            <resource>
                 <directory>src/main/resources</directory>
                 <filtering>false</filtering>
             </resource>
             <resource>
                 <directory>src/main/examples</directory>
                 <filtering>false</filtering>
+                <targetPath>examples</targetPath>
```
m4rkmckenna: Unsure about packaging examples ... 
I think its useful for documentation but would you ever deploy the examples as is?

grkvlt: Correct. Have removed these from the Jars

common/pom.xml
```diff
@@ -46,17 +46,22 @@
                         <configuration>
                             <artifacts>
                                 <artifact>
-                                    <file>${project.basedir}/src/main/resources/docker/docker.bom</file>
+                                    <file>catalog/docker/docker.bom</file>
```
m4rkmckenna: Not sure if we still need `${project.basedir}` I know this was needed when using maven with `-f`

grkvlt: Good point, I think that's right. Will keep it there to be sure, then.

PR #365, opened 2016-10-20
Fix race condition in configureSwarm effector

googlielmo:
- guard critical section with flock

Fixes https://cloudsoft.jira.com/browse/CCS-106

The `swarm-manager-load-balancer` sometimes returns an empty response.
This is due to the fact that several instances of `haproxy` are running, so the request may be routed to an instance with an invalid config.

There is a race condition in the `configureSwarm` effector that is called to re-configure `haproxy` every time there is a change in the swarm node list.

If the effector is invoked simultaneously, e.g. when starting up two nodes, it may result in duplicate instances, as new `haproxy` instances are started with an option to kill and replace the old process, but as they don't know about each other, they both live on.

(A different problem may happen if two effector invocations enter the critical section in the wrong order, but this is not resolvable within the effector itself.)

### DIFF
--- a/common/src/main/resources/common/common.bom
+++ b/common/src/main/resources/common/common.bom
@@ -198,8 +198,11 @@ brooklyn.catalog:
                 balance roundrobin
               EOF
               for endpoint in $(echo ${HAPROXY_ENDPOINTS} | tr ',' ' ') ; do
-                echo "  server manager$((++n)) ${endpoint} maxconn 32 ssl \
-                  ca-file ${RUN_DIR}/ca.pem crt ${RUN_DIR}/cert.pem" >> ${RUN_DIR}/servers.conf
+                if [ "${HAPROXY_PROTOCOL}" == "http" ]; then
+                  echo "  server master$((++n)) ${endpoint} check" >> ${RUN_DIR}/servers.conf
+                else
+                  echo "  server manager$((++n)) ${endpoint} maxconn 32 ssl ca-file ${RUN_DIR}/ca.pem crt ${RUN_DIR}/cert.pem" >> ${RUN_DIR}/servers.conf
+                fi
               done
               OLD_PID=$(cat ${RUN_DIR}/pid.txt)
               haproxy -D -f ${RUN_DIR}/haproxy.conf -f ${RUN_DIR}/servers.conf -st $OLD_PID
--- a/common/src/main/resources/common/common.bom
+++ b/common/src/main/resources/common/common.bom
@@ -140,7 +140,7 @@ brooklyn.catalog:
             HAPROXY_BIND_OPTIONS: $brooklyn:config("haproxy.bind.options")
 
           install.command: |
-            sudo yum install -y gcc make openssl-devel wget
+            sudo yum install -y gcc make openssl-devel wget util-linux
             haproxy_major_minor=$(echo ${HAPROXY_VERSION} | cut -d. -f1,2)
             wget "http://www.haproxy.org/download/${haproxy_major_minor}/src/haproxy-${HAPROXY_VERSION}.tar.gz"
             tar zxf haproxy-${HAPROXY_VERSION}.tar.gz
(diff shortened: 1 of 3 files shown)


grkvlt:
Can we guarantee availability of `flock` in CentOS 7? What about other distros?

neykov:
> (A different problem may happen if two effector invocations enter the critical section in the wrong order, but this is not resolvable within the effector itself.)

Solving this would remove the need for syncing **in** the effector. What do you think about making `InvokeEffectorOnSensorChange` execute the effectors sequentailly?

googlielmo:
@grkvlt good point, as far as it is a "CentOS" system, it _should_ have the `util-linux`package (it's always been part of the minimal install); but it costs us nothing to add an install command to be really sure :-) Will fix.

googlielmo:
@neykov right! it would solve this issue as well.

grkvlt:
Also, do you think we should do the same for the Kubernetes use of the HAProxy entity, which has a similar effector `configureKubernetes`

googlielmo:
@grkvlt yes indeed! 
good point, is it worth generalising the effector in the `haproxy-load-balancer` itself? Also @andreaturli raised a very similar point (that it's probably worth doing it)

common/src/main/resources/common/common.bom
```diff
@@ -172,3 +172,49 @@ brooklyn.catalog:
...
+              cat >> ${RUN_DIR}/servers.conf <<-EOF
+                mode http
+                option httpchk
+              EOF
+              fi
+              cat >> ${RUN_DIR}/servers.conf <<-EOF
+                balance roundrobin
+              EOF
+              for endpoint in $(echo ${HAPROXY_ENDPOINTS} | tr ',' ' ') ; do
+                echo "  server manager$((++n)) ${endpoint} maxconn 32 ssl \
```
grkvlt: This is different for Kubernetes HTTP endpoints:
```
echo "  server master$((++n)) ${endpoint} check" >> ${RUN_DIR}/servers.conf
```

googlielmo: hmm right @grkvlt, I missed that in the refactoring.
How would you tackle this one? a config key to specify options such as `check` or `maxconn 32 ssl`? Or is it deducible from `HAPROXY_PROTOCOL`?
I still think it's worth generalising this entity, but only if feasible.

googlielmo: @grkvlt I've restored the correct k8s functionality in my latest commit; 
very interested in your opinion about this generalisation; see also open questions above

I like the idea of having a load balancer entity that is reusable without much effort (the user only has to provide a sensor called `haproxy.endpoints` and the entity will pick it up at runtime)

common/src/main/resources/common/common.bom
```diff
@@ -172,3 +172,52 @@ brooklyn.catalog:
...
+                else
+                  echo "  server manager$((++n)) ${endpoint} maxconn 32 ssl ca-file ${RUN_DIR}/ca.pem crt ${RUN_DIR}/cert.pem" >> ${RUN_DIR}/servers.conf
+                fi
+              done
+              OLD_PID=$(cat ${RUN_DIR}/pid.txt)
+              haproxy -D -f ${RUN_DIR}/haproxy.conf -f ${RUN_DIR}/servers.conf -st $OLD_PID
+              # wait for the old process to die before leaving the critical section
+              while test -d "/proc/$OLD_PID" 2> /dev/null; do sleep 1; done;
+              flock -u 9
+              ) 9>>/tmp/configureSwarm.lock
```
grkvlt: Should not reference Swarm, and should probably also have entity ID embedded, to deal with running multiple HAProxy instances on the same server (with different ports/protocols, for example) so add an environment variable `ENTITY_ID: $brooklyn:attributeWhenReady("entity.id")` and use `/tmp/configure.${ENTITY_ID}.lock` as the lock file.

common/src/main/resources/common/common.bom
```diff
@@ -172,3 +172,52 @@ brooklyn.catalog:
...
+                if [ "${HAPROXY_PROTOCOL}" == "http" ]; then
+                  echo "  server master$((++n)) ${endpoint} check" >> ${RUN_DIR}/servers.conf
+                else
+                  echo "  server manager$((++n)) ${endpoint} maxconn 32 ssl ca-file ${RUN_DIR}/ca.pem crt ${RUN_DIR}/cert.pem" >> ${RUN_DIR}/servers.conf
+                fi
+              done
+              OLD_PID=$(cat ${RUN_DIR}/pid.txt)
+              haproxy -D -f ${RUN_DIR}/haproxy.conf -f ${RUN_DIR}/servers.conf -st $OLD_PID
+              # wait for the old process to die before leaving the critical section
+              while test -d "/proc/$OLD_PID" 2> /dev/null; do sleep 1; done;
```
grkvlt: Use braces to escape the name, and use lowercase, so `${old_pid}` as this is a local variable; also needs changed in preceding lines

common/src/main/resources/common/common.bom
```diff
@@ -172,3 +172,52 @@ brooklyn.catalog:
           launch.command: |
             haproxy -f haproxy.conf -f servers.conf -c &&
               haproxy -D -f haproxy.conf -f servers.conf -sf $(cat ${PID_FILE})
+
+        brooklyn.initializers:
+        - type: org.apache.brooklyn.core.effector.ssh.SshCommandEffector
+          brooklyn.config:
+            name: configureHaproxy
```
grkvlt: Can we change the effector name to just `configureNodes` to prevent ambiguity about capitalisation of HAProxy in the name?

common/src/main/resources/common/common.bom
```diff
@@ -172,3 +172,52 @@ brooklyn.catalog:
...
+              EOF
+              fi
+              cat >> ${RUN_DIR}/servers.conf <<-EOF
+                balance roundrobin
+              EOF
+              for endpoint in $(echo ${HAPROXY_ENDPOINTS} | tr ',' ' ') ; do
+                if [ "${HAPROXY_PROTOCOL}" == "http" ]; then
+                  echo "  server master$((++n)) ${endpoint} check" >> ${RUN_DIR}/servers.conf
+                else
+                  echo "  server manager$((++n)) ${endpoint} maxconn 32 ssl ca-file ${RUN_DIR}/ca.pem crt ${RUN_DIR}/cert.pem" >> ${RUN_DIR}/servers.conf
```
grkvlt: Minor point, but change both lines to read the same for node names, perhaps `node$((++n))`

googlielmo:
Thanks @grkvlt for the excellent points! 
I've addressed your comments.

neykov:
What would happen if the connection is reset while holding the `flock` and having the `haproxy` launched, but before closing the stream?

googlielmo:
@neykov the lock will be released when the bash process holding it terminates.
Since I'm curious anyway I'll set up a little experiment to test that ;)

googlielmo:
@neykov I found an answer to your question above:
The lock will NOT be released, in fact.
I've addressed this issue in https://github.com/brooklyncentral/clocker/pull/382
cc: @grkvlt @aledsage

PR #368, opened 2016-10-24
Test uses shared security group, no longer does restart test

Graeme-Miller:
CA uses subnet address

### DIFF
--- a/common/catalog/common/ca.bom
+++ b/common/catalog/common/ca.bom
@@ -202,5 +202,5 @@ brooklyn.catalog:
             enricher.targetValue:
               $brooklyn:formatString:
                 - "http://%s:%d"
-                - $brooklyn:attributeWhenReady("host.subnet.address")
+                - $brooklyn:attributeWhenReady("host.address")
                 - $brooklyn:attributeWhenReady("ca.server.port")
--- a/common/catalog/common/ca.bom
+++ b/common/catalog/common/ca.bom
@@ -202,5 +202,5 @@ brooklyn.catalog:
             enricher.targetValue:
               $brooklyn:formatString:
                 - "http://%s:%d"
-                - $brooklyn:attributeWhenReady("host.address")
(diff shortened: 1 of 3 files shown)


swarm/src/test/resources/docker/swarm.tests.bom
```diff
@@ -396,6 +396,7 @@ brooklyn.catalog:
         swarm.port: 4000
         docker.recovery.stabilizationDelay: 10s
         docker.recovery.failOnRecurringFailuresInThisDuration: 5m
+        swarm.sharedsecuritygroup.create: true
```
drigodwin: This is going to be different on differing clouds which we test, do we want to put this here?

Graeme-Miller: I think so. swarm.sharedsecuritygroup.create is a brooklyn.parameter that is defaulted to 'true' in the main swarm blueprint. 

Setting it to true in the test ensure that we test what a user will get by default.

grkvlt: But the point is that on some clouds this default will not work, the user _must_ make a change from the default settings. Or are there going to be specific configurations for those clouds with this set to `false` as well?

grkvlt: @Graeme-Miller there should be tests with both _true_ and _false_ settings for this configuration key, on the appropriate clouds. But, to verify that the default setting work, we should *NOT* be setting it here, for the obvious reason that if the default was ever to change, we would not be "test[ing] what a user will get by default" any more.

swarm/src/main/resources/swarm/swarm.bom
```diff
@@ -446,6 +446,10 @@ brooklyn.catalog:
           default: true
 
       brooklyn.config:
+        name: swarm.create.shared.securitygroup
```
grkvlt: I don't think this will work. It creates four config keys, `name`, `label`, `type`, and `default`.

PR #330, opened 2016-09-30
Adds a multiple master capability

grkvlt:
Load balanced cluster of Kubernetes masters, fronted by HAProxy

PR #371, opened 2016-10-27
Etcd race condition

drigodwin:
Addresses race condition mentioned here - https://github.com/brooklyncentral/clocker/pull/330#issuecomment-251910806

Tested in Bluebox


grkvlt:
Thanks @drigodwin, looks good.

PR #374, opened 2016-11-07
Adds OpenShift Blueprints

grkvlt:
New blueprint for OpenShift clusters, based on Kubernetes template. Includes very basic test case only.


grkvlt:
This method of starting OpenShift Origin is now deprecated, in favour of using the official Red Hat supplied and supported Ansible playbooks, provided by #401

PR #378, opened 2016-11-10
Use local IPs for inter-node communication

neykov:
Using public IPs in Amazon requires the shared security group open to external traffic.

Suggest testing this with the shared security group customizer enabled. If working could update the docs not to require it.

I've tested this with the security groups open to external traffic. If not working with the stock security groups could look at ` --api-servers=${KUBERNETES_URL}` as a potential cause.


grkvlt:
Looks good, have tested on Softlayer.

PR #382, opened 2016-11-14
Prevent haproxy daemon's holding of flock

googlielmo:
Currently fd 9 is associated to flock's lock file in critical section.
 `flock -o 9` was used to ensure that the file itself is closed, while
 the lock is still held by the current process.
 The flock man page explicitly suggests to use `-o` when spawning 
 a child process which should not be holding the lock,
 _but this only works correctly when passing a command_.
 Therefore, after obtaining the lock, the subsequent commands 
 will still have fd 9 available to them.
 When haproxy is started with `-D` it will fork a daemon, and the fd
 will be inherited by the child process.
 Finally, the lock is explicitly released with `flock -u 9`.

 Unfortunately, if any error occurs before releasing the lock (e.g.,
 connection drops) the lock will NOT be released, even if the file is
 closed.
 Inspection of held file locks (with `lslocks`) shows that the original
 parent process ID still holds the lock even after the process has
 been terminated.

 The solution is very simple: redirect fd 9 to `/dev/null` for `haproxy`
 so that the lock is not inherited by the child process.
 In this way the explicit lock release is not necessary any more, and it 
 will happen automatically when the current process exits, either 
 normally or on any error.

_(Edited to clarify why -o does not work as expected)._

### DIFF
--- a/common/catalog/common/haproxy.bom
+++ b/common/catalog/common/haproxy.bom
@@ -110,7 +110,7 @@ brooklyn.catalog:
               ENTITY_ID: $brooklyn:attributeWhenReady("entity.id")
             command: |
               (
-              flock -o 9
+              flock 9
               cat > ${RUN_DIR}/servers.conf <<-EOF
               backend servers
               EOF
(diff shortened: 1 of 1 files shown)


common/catalog/common/haproxy.bom
```diff
@@ -110,7 +110,7 @@ brooklyn.catalog:
               ENTITY_ID: $brooklyn:attributeWhenReady("entity.id")
             command: |
               (
-              flock -o 9
+              flock 9
```
aledsage: Why is the lockfile called "9"? Is that some naming convention that I'm not aware of?

aledsage: Ah, `9` is the file descriptor? See http://stackoverflow.com/questions/8866175/preventing-lock-propagation

aledsage:
Would we still have a problem (i.e. the lock not released) if something went wrong with the connection, causing the remaining ssh commands to not execute because the session had closed prematurely? Even if that is the case, I'm happy for us to tackle this incrementally.

I'll leave it to someone who knows more about `flock` to review this :-)
@grkvlt @geomacy @neykov perhaps one of you could review please?

common/catalog/common/haproxy.bom
googlielmo: Excellent SO discussion @aledsage, I overlooked it. 
There are interesting variations on the same theme there, but the conclusion appears the same: make sure that the file descriptor is not propagated to the child process.

PR #384, opened 2016-11-18
Enable testing in clouds that do not require a shared security group

m4rkmckenna:


neykov:
Makes sense. What do you think about adding code to `SharedLocationSecurityGroupCustomizer` to skip clouds it doesn't support by default (read OpenStack)?
Adding the option is obviously a painful workaround.

PR #385, opened 2016-11-18
Use integer type for port parameters

googlielmo:


andreaturli:
lgtm, easiest fix, probably.

out of curiosity, have you got to the bottom of the problem? why `type: port` is not coerced correctly?

googlielmo:
@andreaturli not yet... looking into it

andreaturli:
lgtm!
Thanks

PR #386, opened 2016-11-18
Fix `swarm.port` usage

googlielmo:
- use `type: integer` for the parameter
- usage of `swarm.port` as config, not as a sensor

### DIFF
--- a/common/catalog/docker/docker.bom
+++ b/common/catalog/docker/docker.bom
@@ -192,7 +192,7 @@ brooklyn.catalog:
           label: "Docker Port"
           description: |
             The TCP port for Docker to listen on
-          type: port
+          type: integer
           default: 2376
         - name: docker.bindaddress
           label: "Docker Bind Address"
--- a/common/catalog/docker/docker.bom
+++ b/common/catalog/docker/docker.bom
@@ -236,26 +236,24 @@ brooklyn.catalog:
             enricher.suppressDuplicates: false
             enricher.triggerSensors:
               - $brooklyn:sensor("host.address")
-              - $brooklyn:sensor("docker.port")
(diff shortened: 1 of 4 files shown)


Graeme-Miller:
Looks good, but there is a few places we use:
attributeWhenReady("ca.server.port")

We'll need to change those to config("ca.server.port")

googlielmo:
@Graeme-Miller I've addressed your comment above

common/catalog/docker/docker.bom
```diff
@@ -192,7 +192,7 @@ brooklyn.catalog:
           label: "Docker Port"
           description: |
             The TCP port for Docker to listen on
-          type: port
+          type: integer
```
grkvlt: Although this is changed, none of the `attributeWhenReady("docker.port")` uses have been updated

common/catalog/docker/docker.bom
```diff
@@ -271,14 +269,12 @@ brooklyn.catalog:
           brooklyn.config:
             uniqueTag: docker-bind-url-generator
             enricher.suppressDuplicates: false
-            enricher.triggerSensors:
-              - $brooklyn:sensor("docker.port")
```
grkvlt: This isn't going to work any more, as there are no incoming sensors. We will need something like a `StaticSensor` initializer instead.

grkvlt: OK, this is incorrect - a `Transformer` will execute when it is first created, so this will run and use the two config key values. It just won't run again, but this is not required.

grkvlt:
LGTM @googlielmo since the transformer issue is not relevant

googlielmo:
Cool, thanks @grkvlt

PR #391, opened 2016-12-07
Change load balancer URI to use host.name for AWS

grkvlt:
Required because AWS VPCs set the `host.address` sensor to a VPC local IP address. The `host.name` will resolve correctly to the public IP externally and the VPC IP internally, so use this for the reported `kubernetes.url` sensor that is propagated to the application root as `main.uri` for the dashboard UI.


andreaturli:
+1

looks useful from my tests as well

PR #392, opened 2016-12-07
HAProxy mode fix and add stats option

grkvlt:
Fix setting of proxy mode for HTTP and add options to enable statistics page. Testing requirements same as master, stats page disabled by default.


drigodwin:
I get this when I run this entity:
```
http frontend 'www-http' (haproxy.conf:11) tries to use incompatible tcp backend 'servers' (servers.conf:1) as its default backend (see 'mode').
```

PR #395, opened 2016-12-12
Remove external resources

grkvlt:
**Important** This is required to let the Kubernetes catalog entries load when there is no external (public Internet) connectivity available to AMP.

Updates `iconUrl` to point at local files with `classpath://` URIs and adds the icons as resources. Copies the [brooklyn-dns](https://github.com/brooklyncentral/brooklyn-dns) BOM file to the target Jar so that it can be imported from a `classpath://` URI also.


andreaturli:
Why from http to classpath? Performance?

andreaturli:
Ah no maybe it's for weird networks where those resources may be nor reachable

grkvlt:
So that the BOM file can be loaded onto a Brooklyn server running somewhere that does not have Internet access, such as many large enterprises.

drigodwin:
@grkvlt, have you checked if we have the rights to incorporate all these images in clocker? Do they meet whatever terms of use they have?

grkvlt:
@drigodwin good point, will review the terms of use and update if required

grkvlt:
See [guidelines](https://github.com/kubernetes/kubernetes/blob/master/logo/usage_guidelines.md) for Kubernetes, also others. Will document these links on the Wiki here.

grkvlt:
@drigodwin have added license details for icons

PR #396, opened 2016-12-13
Add loginUser to centos7 entity

drigodwin:
This is short term amendment until [JCLOUDS-1165](https://issues.apache.org/jira/browse/JCLOUDS-1165) is addressed. Currently unless `loginUser: centos` is specified in the location of a CentOS image, Jclouds fails to provision the machine.


drigodwin:
This is unfortunately not going to work because the RightScale image which Brooklyn provisions by default when launching with CentOS selected has the default use of `root`.

PR #401, opened 2017-01-13
OpenShift Origin Deployment with Ansible

grkvlt:
Uses the Red Hat [`openshift-ansible`](https://github.com/openshift/openshift-ansible) repository to deploy OpenShift Origin using the provided Playbooks. Creates a set of machines that are then used to populate an inventory file as the source configuration for Ansible. This deployment mechanism is also easily convertible to create an OpenShift Container Platform for enterprise environments.

Replaces #374 which uses a deprecated deployment mechanism for Origin.

**Note** Because of https://github.com/openshift/openshift-ansible/issues/3111 we have to use Ansible 2.2.0.0 but once this is fixed we should look at reverting f6b0e11


openshift/catalog/openshift/openshift.bom
```diff
@@ -0,0 +1,720 @@
...
+              - type: org.apache.brooklyn.enricher.stock.Transformer
+                brooklyn.config:
+                  enricher.triggerSensors:
+                    - host.address
+                  enricher.targetSensor: $brooklyn:sensor("main.uri")
+                  enricher.targetValue:
+                    $brooklyn:formatString:
+                      - "https://%s:8443/"
+                      - $brooklyn:attributeWhenReady("host.address")
+              - type: io.cloudsoft.amp.ui.common.enricher.AmpExternalUiModuleEnricher
```
robertgmoss: are we using this? if not can it be removed - deploying with vanilla Brooklyn requires it to be removed

grkvlt:
Closed, to allow development in a separate repository

PR #408, opened 2017-03-14
Removed check for Prometheus UI being Reachable. Depending on where t…

Graeme-Miller:
…his is deployed, it may not be reachable

### DIFF
--- a/kubernetes/tests/kubernetes/kubernetes.tests.bom
+++ b/kubernetes/tests/kubernetes/kubernetes.tests.bom
@@ -116,17 +116,7 @@ brooklyn.catalog:
           command: |
             kubectl get deployments | grep workload- | wc -l
           assertOut:
-            contains: '0'
-        - type: org.apache.brooklyn.test.framework.TestHttpCall
-          name: "14. Check Prometheus UI is Reachable"
-          url:
-            $brooklyn:formatString:
-              - "http://%s:%s"
-              - $brooklyn:entity("prometheus").attributeWhenReady("host.address")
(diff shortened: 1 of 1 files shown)


kubernetes/tests/kubernetes/kubernetes.tests.bom
```diff
@@ -116,17 +116,7 @@ brooklyn.catalog:
...
-          name: "14. Check Prometheus UI is Reachable"
-          url:
-            $brooklyn:formatString:
-              - "http://%s:%s"
-              - $brooklyn:entity("prometheus").attributeWhenReady("host.address")
-              - $brooklyn:entity("prometheus").attributeWhenReady("kubernetes.service.port")
-          applyAssertionTo: status
-          assert:
-          - equals: 200
+            contains: '0'
```
m4rkmckenna: Why not change this to `TestSshCommand` and use curl

grkvlt: Because this is _technically_ more correct. We are testing HTTP(S) access, not SSH/shell. And also, these tests may be run against a deployed Kubernetes cluster that cannot be SSHed to, for example.

Graeme-Miller:
Fixed in this PR instead: https://github.com/brooklyncentral/clocker/pull/414

kubernetes/tests/kubernetes/kubernetes.tests.bom
Graeme-Miller: I don't think we want to run these tests against a Kubernetes cluster that cannot be SSHed to.

I think it is more likely we run them against clouds on which we don't open the high value ports, so I agree with Marks suggestion.

PR #423, opened 2017-05-16
Fix DNS and update Kubernetes test id

grkvlt:
Use `brooklyn-dns` version 0.1.3 which fixes concurrency issues and rename the root Kubernetes test catalog entity for eventual use with OpenShift deployments

### DIFF
--- a/kubernetes/tests/kubernetes/kubernetes.tests.bom
+++ b/kubernetes/tests/kubernetes/kubernetes.tests.bom
@@ -1,21 +1,26 @@
 brooklyn.catalog:
   items:
-  - https://raw.githubusercontent.com/brooklyncentral/common-catalog-utils/master/common-tests/src/main/resources/commontests/common.tests.bom
-  - id: kubernetes-cluster-tests
-    version: "2.1.0-SNAPSHOT" # CLOCKER_VERSION
-    itemType: template
-    iconUrl: https://twitter.com/kubernetesio/profile_image?size=original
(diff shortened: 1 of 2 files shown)


kubernetes/pom.xml
```diff
@@ -14,6 +14,10 @@
     <artifactId>clocker-kubernetes</artifactId>
     <packaging>bundle</packaging>
 
+    <properties>
+        <brooklyn-dns.version>0.1.3</brooklyn-dns.version> <!-- BROOKLYN_DNS_VERSION -->
```
m4rkmckenna: Move to root pom

grkvlt: Moved, thanks

kubernetes/tests/kubernetes/kubernetes.tests.bom
```diff
@@ -1,21 +1,26 @@
...
+        services:
+          - type: kubernetes-cluster-template
+            id: k8s-cluster
+          - type: kubernetes-smoke-tests
+            id: k8s-smoke-tests
+
+    - id: kubernetes-smoke-tests
+      type: org.apache.brooklyn.test.framework.TestCase
+      name: K8S Smoke Tests
+      brooklyn.config:
```
m4rkmckenna: Should this config block be removed from here and moved into the `kubernetes-cluster-tests` block

grkvlt: Yep

PR #435, opened 2017-08-31
Update tests.no-shared-security-group.bom

Graeme-Miller:
K8s tests sometimes fail due to the deployment tests causing a resize. After the deployment tests the resize tests would assume a particular cluster size, but it would be wrong.

Increase autoscaler.resizeUpStabilizationDelay from the default of 30s to 60s.

re-arranged k8s tests so the resizing tests happen before deployment tests.

### DIFF
--- a/kubernetes/tests/kubernetes/kubernetes.tests.bom
+++ b/kubernetes/tests/kubernetes/kubernetes.tests.bom
@@ -17,9 +17,9 @@ brooklyn.catalog:
           targetId: k8s-cluster
         brooklyn.children:
         - type: assert-up-and-running-initial
-          name: "01. K8S cluster up and running"
+          name: "1. K8S cluster up and running"
         - type: assert-reachable
-          name: "02. K8S UI Reachable"
+          name: "2. K8S UI Reachable"
           brooklyn.config:
             endpointSensor: main.uri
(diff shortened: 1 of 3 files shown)


kubernetes/tests/kubernetes/kubernetes.tests.bom
```diff
@@ -23,35 +23,81 @@ brooklyn.catalog:
           brooklyn.config:
             endpointSensor: main.uri
             timeout: 5m
+        - type: org.apache.brooklyn.test.framework.TestSensor
+          name: "3. Size of kubernetes master cluster is 3"
```
duncangrant: This may be the most nit-picky review point ever made but now the numbering goes 01 02 3 4.  I won't take it personally if you ignore this.

kubernetes/tests/kubernetes/kubernetes.tests.bom
```diff
@@ -23,35 +23,81 @@ brooklyn.catalog:
           brooklyn.config:
             endpointSensor: main.uri
             timeout: 5m
+        - type: org.apache.brooklyn.test.framework.TestSensor
```
duncangrant: use assert-sensor?

PR #437, opened 2017-09-25
Rename main bom file for easy import via brooklyn.libraries

tbouron:
Rename `clocker.bom` to `catalog.bom` to be able to include this bundle via `brooklyn.libraires`:

```yaml
brooklyn.catalog:
  brooklyn.libraires:
  - mvn:io.brooklyn.clocker/clocker-parent
```


tbouron:
That's not the right way of doing this. Will push another PR shortly

PR #438, opened 2017-09-26
Update packaging config to allow installation via brooklyn.libraries

tbouron:
To install a bundle via a bom file and brooklyn.libraries, you need to have:

- a valid OSGi bundle (was already there)
- a `catalog.bom` at the root of the bundle, which this PR adds

It also reshuffle a bit the bom files to make internal boms use` brooklyn.libraries` when necessary


common/catalog/catalog.bom
```diff
@@ -22,4 +23,14 @@ brooklyn.catalog:
       iconUrl: classpath://io.brooklyn.clocker.common:icons/haproxy.png
       item:
         services:
-          - type: haproxy-load-balancer
\ No newline at end of file
+          - type: haproxy-load-balancer
+
+    - id: docker-engine-template
```
m4rkmckenna: Can you remove this template ... 
IMO its just wrapping the docker-engine entity adding no extra value

kubernetes/catalog/kubernetes/kubernetes.bom
```diff
@@ -1,11 +1,12 @@
...
+  brooklyn.libraries:
+  - mvn:io.brooklyn.clocker/clocker-common
+  - mvn:io.brooklyn.etcd/brooklyn-etcd
+
   items:
-  - classpath://io.brooklyn.etcd.brooklyn-etcd:etcd/etcd.bom
-  - classpath://io.brooklyn.clocker.common:docker/docker.bom
-  - classpath://io.brooklyn.clocker.common:common/haproxy.bom
-  - classpath://io.brooklyn.clocker.kubernetes:kubernetes/brooklyn-dns-etc-hosts-generator.bom
+  - classpath://kubernetes/brooklyn-dns-etc-hosts-generator.bom
```
m4rkmckenna: The way this bundle makes it way into clocker seems hacky now https://github.com/brooklyncentral/clocker/blob/master/kubernetes/pom.xml#L39-L56

Can you update it as part of this PR please

common/catalog/catalog.bom
tbouron: Sure, I'll remove it

tbouron:
@m4rkmckenna Comments addressed. Although, this now requires https://github.com/brooklyncentral/brooklyn-dns/pull/2 and we probably need to build and push it to maven central before this PR

Graeme-Miller:
Merged. Marks comments have been addressed

PR #444, opened 2017-10-31
Add custom liquid tags for tabs

tbouron:
This add custom liquid tag to handle the tab. However, custom plugins cannot be ran automatically on GitHub pages therefore the HTML page needs to be built locally first, then pushed to the `gh-pages` branch.

That is why this PR is again a `website` branch. This branch is a clone of the current `gh-pages` and includes a brand new `publish.sh`. This script will build, and push the HTML pages onto the `gh-pages` branch upon execution.


publish.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+GIT_REPO=https://github.com/tbouron/clocker
```
drigodwin: This probably isn't the correct repo here?

tbouron: Ooops, you are right, that was for my test. Updated @drigodwin

PR #445, opened 2018-01-17
Separate Template from entity

duncangrant:
Template not always useful so allows someone to only add entity.

The swarm template currently only deploys an out of date swarm so this makes it easier to remove from the catalog.

### DIFF
--- a/swarm/catalog/swarm/swarm.bom
+++ b/swarm/catalog/swarm/swarm.bom
@@ -10,7 +10,7 @@ brooklyn.catalog:
     name: "Docker Swarm with Discovery and CA"
     description: |
       Creates a Swarm of Docker engines, of configurable initial size, using an Etcd cluster for discovery.
-    itemType: template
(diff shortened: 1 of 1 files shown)


swarm/catalog/swarm/swarm.bom
```diff
@@ -6,15 +6,6 @@ brooklyn.catalog:
   - mvn:io.brooklyn.etcd/brooklyn-etcd/2.7.0-SNAPSHOT # BROOKLYN_ETCD_VERSION
 
   items:
-  - id: docker-swarm-template
```
aledsage: I think we need to leave this in case someone's persisted state references it.

However, we should change it to `itemType: entity` - that will look a bit weird (given the entity name), but it's not for the swarm bundle to say what should go in "quick launch". That should be up to the person who is configuring brooklyn to decide which of their bundle's entities should be "quick launch" items.

PR #450, opened 2018-06-19
Add stabilisation delay

duncangrant:
Adds a stabilisation delay config parameter.

Increases the stabilisation delay and upper bound cpu for scaling in tests. The tests were sometimes scaling up during initial few tests possibly due to killing a master node.

### DIFF
--- a/kubernetes/catalog/kubernetes/kubernetes.bom
+++ b/kubernetes/catalog/kubernetes/kubernetes.bom
@@ -186,6 +186,13 @@ brooklyn.catalog:
             will automatically be added. The default is 0.95 or 95%
           type: double
           default: 0.95
+        - name: kubernetes.resizeUpStabilizationDelay
+          label: "Kubernetes Resize Stabilization Delay"
+          description: |
+            The time that the average cluster cpu must be above the threshold before another
+            node will automatically be added.
+          type: double
+          default: 30s
         - name: kubernetes.cluster.name
           label: "Kubernetes Cluster Name"
           type: string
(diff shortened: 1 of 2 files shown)


kubernetes/catalog/kubernetes/kubernetes.bom
```diff
@@ -186,6 +186,13 @@ brooklyn.catalog:
             will automatically be added. The default is 0.95 or 95%
           type: double
           default: 0.95
+        - name: kubernetes.resizeUpStabilizationDelay
+          label: "Kubernetes Resize Stabilization Delay"
+          description: |
+            The time that the average cluster cpu must be above the threshold before another
+            node will automatically be added.
+          type: double
```
tbouron: The type is `double` but the default value is `30s`, it doesn't seem to match. Shouldn't it be `java.time.Duration` instead?

kubernetes/tests/kubernetes/tests.default.bom
```diff
@@ -27,5 +27,6 @@ brooklyn.catalog:
           kubernetes.minRam: 2000
           kubernetes.minCores: 1
           autoscaler.resizeUpStabilizationDelay: 60s
+          kubernetes.scaling.cpu.limit: 2.0
```
tbouron: Is that also part of this PR? Not sure what it does

duncangrant: Increases the threshold that will trigger the up scaling.  I don't know exactly what this represents but in general a high number seems to be about 1.2 so this should stop scaling.  While making the resizeUpStabalization delay is probably enough I couldn't really tell from just sensor info.

kubernetes/catalog/kubernetes/kubernetes.bom
duncangrant: Oops - thought I'd copied that from the autoscaler policy but clearly I did not - I'll fix.

PR #451, opened 2018-10-02
qa tests: reboot machine, rather than replace

aledsage:
Motivated by https://issues.apache.org/jira/browse/BROOKLYN-603, which can cause this test to fail if a cloud-cleaner deletes the keypair in the 10 minutes between the VM being initially created and then the VM replaced by the call to `restart(restartMachine: true)`.

Instead, a better test is that the VM survives reboot. I think that is more useful to the user, rather than replacing the VM entirely.

### DIFF
--- a/common/tests/docker/docker.tests.bom
+++ b/common/tests/docker/docker.tests.bom
@@ -40,8 +40,12 @@ brooklyn.catalog:
           name: "TEST-03 start process"
         - type: test-restart-process
           name: "TEST-04 restart process"
-        - type: test-restart-machine
-          name: "TEST-05 restart machine"
+        - type: test-case
+          name: "TEST-05 reboot machine"
+          brooklyn.children:
+            - type: ssh-cmd-restart
+            - type: assert-failed
+            - type: assert-up
         - type: test-reachable
           name: "TEST-06 check service is reachable"
           brooklyn.config:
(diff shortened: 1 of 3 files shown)


common/tests/docker/docker.tests.bom
```diff
@@ -40,8 +40,12 @@ brooklyn.catalog:
...
         - type: test-restart-process
           name: "TEST-04 restart process"
-        - type: test-restart-machine
-          name: "TEST-05 restart machine"
+        - type: test-case
+          name: "TEST-05 reboot machine"
+          brooklyn.children:
+            - type: ssh-cmd-restart
+            - type: assert-failed
+            - type: assert-up
```
tbouron: Don't we also want a `type: assert-running` ?

tbouron (commented):
@aledsage LGTM, just one tiny comment

aledsage:
PR comment addressed.

Travis failure is:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:3.0.0-M1:install (default-install) on project clocker-feature: NoFileAssignedException: The packaging plugin for this project did not assign a main file to the project but it has attachments. Change packaging to 'pom'. -> [Help 1]
```
This looks familiar. I think that @kemitix hit that somewhere else, and rolled back `maven-install-plugin` to a version older than 3.0.0-M1 to fix it. I've therefore added that fix as well.

PR #452, opened 2018-10-19
Hold back maven-install-plugin to 2.5.2 MINSTALL-151

kemitix:
[ERROR] Failed to execute goal
org.apache.maven.plugins:maven-install-plugin:3.0.0-M1:install (default-install)
on project clocker-features: NoFileAssignedException: The packaging plugin for
this project did not assign a main file to the project but it has attachments.
Change packaging to 'pom'. -> [Help 1]

The maven-install-plugin 3.0.0-M1 doesn't handle modules with no primary
artefacts, but that still has attached artefacts.

https://issues.apache.org/jira/browse/MINSTALL-151

Fixes https://jenkins.cloudsoftdev.net/job/builds/job/brooklyn-central/job/clocker/445/

PR #453, opened 2018-10-19
Fix version for maven-deploy-tool

tbouron:
Fix the same issue as #452 that also affects `maven-deploy-plugin`


kemitix (approved):
LGTM @tbouron

